PR #8, opened 2022-01-12
Post Folio records from comma-seperated array of json objects

jgreben:
Adds a comma to the end of all the converted folio instance files and then moves work files to the archive folder, then read files to post from a temporary comma-seperated json file.


dags/folio_post.py
```diff
@@ -52,14 +52,9 @@ def _post_to_okapi(**kwargs):
...
 
-# Maybe check for empty files here...
-# Also each json file is just an object on each line, not an array
-# Parse each json file and add to (comma separated) array that gets posted as the payload on line 45.
 def post_folio_instance_records(**kwargs):
     """Creates new records in FOLIO"""
-    inventory_records = [fo.read_text() for fo in pathlib.Path(
-        "/opt/airflow/migration/results").glob("folio_instance_*.json"
-        )]
+    instance_records = pathlib.Path('/tmp/instances.json')
```
jermnelson, line 57: I think we still need to read the file contents of `instance.json` so this line can be changed to `instance_records = pathlib.Path('/tmp/instances.json').read_text()`

jermnelson (changes requested):

PR #30, opened 2022-02-02
Makes folio login a separate task.

### LINKED ISSUE #19: Create FOLIO Login Task
When #18 is merged, we are calling `FolioLogin` function twice for each of the holdings and instances POST calls. We should following the pattern used in Sinopia's [ils-middleware](https://github.com/LD4P/ils-middleware/blob/aee729117a1607ba9fe18230a1e565b398666cba/ils_middleware/dags/stanford.py#L161-L210). We could then pass in the JWT to each of the post function calls.

shelleydoljack:
Closes #19

### DIFF
--- a/dags/bib_records.py
+++ b/dags/bib_records.py
@@ -150,7 +150,7 @@ with DAG(
 
     with TaskGroup(group_id="post-to-folio") as post_to_folio:
 
-        folio_login = PythonOperator(task_id="folio_login", python_callable=folio_login)
+        login = PythonOperator(task_id="folio_login", python_callable=folio_login)
 
         post_instances = PythonOperator(
             task_id="post_to_folio_instances",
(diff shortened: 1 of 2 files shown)


dags/bib_records.py
```diff
@@ -149,6 +150,8 @@ def move_marc_files(*args, **kwargs) -> list:
 
     with TaskGroup(group_id="post-to-folio") as post_to_folio:
 
+        folio_login = PythonOperator(task_id="folio_login", python_callable=folio_login)
```
jermnelson, line 153: To reduce confusion, could you rename the `PythonOperator` instance  `folio_login` to differentiate from the import function `folio_login`?

jermnelson (changes requested):
Looks good; just a small request to rename a task instance.

jermnelson (approved):

PR #43, opened 2022-02-23
Install Python dependencies, run tests and lint

jgreben:


jermnelson (approved):
Approved but could you remove the __pycache__ files from the PR?

Issue #48, opened 2022-02-28
Test functions in FOLIO helpers.py 


jermnelson:
Test the following functions in the `helpers` module:

- [x] **archive_artifacts** -- relatively easy to test, need to mock dag and use tmp_file for airflow path. Also, need to change the `archive_artifacts` function to look in `/opt/airflow/migration/results` directory for files to archive. Fixed with PR #52 
- [ ] **move_marc_files_check_csv** -- harder to test, need to mock XCOM as it sets a `marc_only` key for use in downstream tasks
- [ ] **process_records** -- moderate test difficultly, mostly around mocking file system.  Need to include a change to the `process_records` function to pass in parameter for `airflow` path (see `archive_artifacts` for example)
- [ ] **tranform_csv_to_tsv** -- moderate test difficulty, as with the `process_records` should modify so that we can pass in a parameter the location to `/opt/airflow` (see above)

**NOTE** All of these functions involve interactions with the filesystem and will require use of pytest's [tmp_file](https://docs.pytest.org/en/6.2.x/tmpdir.html)

PR #52, opened 2022-03-01
Test for archive_artifacts function

### LINKED ISSUE #51: Use correct directory to archive DAG Run JSON artifacts
We should be looking for the Bib, Holdings, and Item transformation JSON artifacts in the `/opt/airflow/migration/results` directory at: https://github.com/sul-dlss/folio-airflow/blob/e11ca3333877a2943490cee9583c563ba5fdf06e/plugins/folio/helpers.py#L24

Mentioned as part of the work https://github.com/sul-dlss/folio-airflow/issues/48

jermnelson:
FIXES #51 part of #48


shelleydoljack (approved):
Thanks for this. I wasn't sure all that mocking was required but I see it is.

PR #55, opened 2022-03-08
Adds tests for the rest of the helper functions

### LINKED ISSUE #48: Test functions in FOLIO helpers.py

shelleydoljack:
Closes #48

### DIFF
--- a/dags/bib_records.py
+++ b/dags/bib_records.py
@@ -21,7 +21,7 @@ from plugins.folio.helpers import (
     move_marc_files_check_csv,
     process_marc,
     process_records,
-    tranform_csv_to_tsv,
+    transform_csv_to_tsv,
 )
 from plugins.folio.holdings import (
     run_holdings_tranformer,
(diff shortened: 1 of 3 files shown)


jermnelson (approved):
Looks good!

PR #60, opened 2022-03-10
Adds Data Issue logging and display in FOLIO plugin

### LINKED ISSUE #58: Enable logging level 26 for DATA Issue generation
The [folio_migration_tools](https://github.com/FOLIO-FSE/folio_migration_tools/blob/bf2f3afdae7a0cd93686ab86c4ecb06eb4946fbc/migration_tools/helper.py#L64) logs DATA Issues at a custom log level of 26. In the folio_plugin, we disable the tool's custom logging in the transformers because of a race condition with Airflow. We need to update this custom logging to capture the DATA issues so that they can be displayed in the Airflow UI FOLIO Reports.

jermnelson:
Fixes #58 

Data Issues now appear as a report and looks like:
Screen Shot 2022-03-09 at 5 13 06 PM

TODO:
- [x] Test `_extract_dag_run_id` for data issues file
- [x] Test `setup_data_logging`

### DIFF
--- a/plugins/folio/helpers.py
+++ b/plugins/folio/helpers.py
@@ -8,6 +8,7 @@ import pymarc
 import requests
 
 from airflow.models import Variable
+from migration_tools.migration_tasks.migration_task_base import LevelFilter
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


plugins/folio/helpers.py
```diff
@@ -188,6 +189,25 @@ def process_records(*args, **kwargs) -> list:
     return len(records)
 
 
+def setup_data_logging(transformer):
+    def transformer_data_issues(transformer, message, *args, **kwargs):
+        transformer._log(DATA_ISSUE_LVL_NUM, message, args, **kwargs)
+
+    # Set DATA_ISSUE logging levels
+    DATA_ISSUE_LVL_NUM = 26
+    logging.addLevelName(DATA_ISSUE_LVL_NUM, "DATA_ISSUES")
```
jgreben, line 198: Should we use or add `getLevelName(level)` for the text instead of "DATA_ISSUES"? Or maybe they are all just errors anyway?

jgreben (commented):
Looks good, just wondering if it's worth adding info about the log level to the data-frame.

plugins/folio/helpers.py
jermnelson: Since  `DATA_ISSUE_LVL_NUM` is a custom log level, calling `getLevelName` just returns a default label. This use of `addLevelName` just makes it more explicit that the log level is for data issues.  

```python
>>> import logging
>>> DATA_ISSUE_LVL_NUM = 26
>>> logging.getLevelName(DATA_ISSUE_LVL_NUM)
'Level 26'
>>> logging.addLevelName(DATA_ISSUE_LVL_NUM, 'DATA_ISSUES')
>>> logging.getLevelName(DATA_ISSUE_LVL_NUM)
'DATA_ISSUES'
```

jgreben (approved):

PR #62, opened 2022-03-11
Disabled Flower service

jermnelson:
Once this PR [1165](https://github.com/mher/flower/pull/1165) is merged to fix security issue with an old version of jquery; we can renabled flower if we run into problems with Celery in our Airflow environment.


shelleydoljack (approved):
👍

PR #75, opened 2022-04-06
Remediation DAG for FOLIO Records Errors

### LINKED ISSUE #45: Remediation DAG for failed Okapi Records
When PR #46 is merged, a failed batch POST to Okapi for inventory instances, holdings, and items will generate a JSON file containing all of the respective UUIDs of the failed record load. This new DAG could be triggered by a completion of the `symphony_marc_import` DAG that would iterate through each of these files, extract the ID and the record, and attempt to POST them individually. 

The order should instance, holdings, and items. 

Open question:
1. How should we handle subsequent failures?

jermnelson:
Fixes #45 

New `fix_failed_record_loads` DAG is triggered after `symphony_marc_import` DAG Run. The `fix_failed_record_loads` tries to re-post each record in `errors-instance-storage--{error-code}-{dag-run.id}.json`, `errors-holdings-storage-{error-code}-{dag-run.id}.json`, and `errors-items-storage-{error-code}-{dag-run.id}.json` created during the `symphony_marc_import` DAG Run. 

TODO:
- [x] Increase test coverage of `remediate.py` module to at least 80%

### DIFF
--- a/dags/bib_records.py
+++ b/dags/bib_records.py
@@ -10,6 +10,7 @@ from airflow import DAG
 
 from airflow.operators.dummy import DummyOperator
 from airflow.operators.python import BranchPythonOperator, PythonOperator
+from airflow.operators.trigger_dagrun import TriggerDagRunOperator
 from airflow.sensors.filesystem import FileSensor
 from airflow.utils.task_group import TaskGroup
 from airflow.models import Variable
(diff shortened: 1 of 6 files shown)


jgreben (approved):
Looks good. I am just wondering in general why records would be failing to post to begin with, and if it has something to do with the records themselves, would trying to re-post them later do any good? What is the provenance of sul-dlss/libsys-airflow/issues/45?

jermnelson:
> Looks good. I am just wondering in general why records would be failing to post to begin with, and if it has something to do with the records themselves, would trying to re-post them later do any good? What is the provenance of [/issues/45](https://github.com/sul-dlss/libsys-airflow/issues/45)?

The context for #45  are the hrid collusions errors we are/were occasionally seeing when trying to post multiple records in a batch. If the batch fails, none of the records are posted so this DAG tries to post each record individually and log the result.

jgreben:
Ok thanks for reminding me! In that case, reposting will help.

PR #83, opened 2022-04-15
Use TSV exports from Symphony

### LINKED ISSUE #82: Refactor to support incoming TSV files with headers
We now are generating tsv files with the proper headers and do not need much of the functionality in the `transform_csv_to_tsv` function in the `helpers.py`.

jermnelson:
FIXES #82

Refactor to support Symphony TSV file exports that have field headers.

Also adds a transform for stripping out extra spaces in BARCODE .

### DIFF
--- a/plugins/folio/helpers.py
+++ b/plugins/folio/helpers.py
@@ -40,7 +40,7 @@ def archive_artifacts(*args, **kwargs):
         logger.info("Moved {artifact} to {target}")
 
 
-def move_marc_files_check_csv(*args, **kwargs) -> str:
+def move_marc_files_check_tsv(*args, **kwargs) -> str:
     """Moves MARC files to migration/data/instances, sets XCOM
     if csv is present"""
     task_instance = kwargs["task_instance"]
@@ -54,9 +54,9 @@ def move_marc_files_check_csv(*args, **kwargs) -> str:
     if not marc_path.exists():
         raise ValueError(f"MARC Path {marc_path} does not exist")
 
-    # Checks for CSV file and sets XCOM marc_only if not present
+    # Checks for TSV file and sets XCOM marc_only if not present
     csv_path = pathlib.Path(
-        f"{airflow}/{source_directory}/{marc_path.stem}.csv"
+        f"{airflow}/{source_directory}/{marc_path.stem}.tsv"
     )
     marc_only = True
     if csv_path.exists():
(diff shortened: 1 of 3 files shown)


plugins/folio/helpers.py
```diff
@@ -219,30 +219,40 @@ def transformer_data_issues(transformer, message, *args, **kwargs):
...
+
+def transform_move_tsvs(*args, **kwargs):
+    airflow = kwargs.get("airflow", "/opt/airflow")
+    column_transforms = kwargs.get("column_transforms", [])
+
+    csv_paths = _get_tsv(**kwargs)
+
+    path_names = [f"{path.name}.tsv" for path in csv_paths]
+
+    if len(csv_paths) < 1:
```
jgreben, line 250: tsv instead of csv here?

jgreben: also on line 252?

plugins/folio/helpers.py
```diff
@@ -219,30 +219,40 @@ def transformer_data_issues(transformer, message, *args, **kwargs):
...
-    df = pd.read_csv(csv_path, names=column_names)
 
-    # Performs any transformations to values
-    for transform in column_transforms:
-        column = transform[0]
-        function = transform[1]
-        df[column] = df[column].apply(function)
-    tsv_path = pathlib.Path(f"{airflow}/migration/data/items/{marc_stem}.tsv")
-    df.to_csv(tsv_path, sep="\t", index=False)
+    for path in csv_paths:
```
jgreben, line 255: tsv?

plugins/folio/helpers.py
```diff
@@ -219,30 +219,40 @@ def transformer_data_issues(transformer, message, *args, **kwargs):
...
+def _apply_processing_tsv(tsv_path, airflow, column_transforms):
+    df = pd.read_csv(tsv_path, sep="\t")
+    # Performs any transformations to values
+    for transform in column_transforms:
+        column = transform[0]
+        if column in df:
+            function = transform[1]
+            df[column] = df[column].apply(function)
+    new_tsv_path = pathlib.Path(f"{airflow}/migration/data/items/{tsv_path.name}")
+    df.to_csv(new_tsv_path, sep="\t", index=False)
```
jgreben, line 231: are we actually converting to csv here? I thought this could be removed because it is natively tsv?

jgreben (changes requested):
I had some confusion about tsv vs csv in this PR. Maybe you can walk me through this one?

plugins/folio/helpers.py
jermnelson: The pandas `to_csv` function with the `sep="\t"` parameter saves the file as tsv file. Since we're potentially doing some manipulation of the data in lines 225-229, we still need to make this call to update the tsv.

plugins/folio/helpers.py
jermnelson: Done, pushed up new commit.

plugins/folio/helpers.py
jermnelson: Done.

plugins/folio/helpers.py
jgreben: gotcha.

jgreben (approved):

PR #85, opened 2022-04-19
HRIDs based on CATKEYs

jermnelson:
Generates HRID for instances, holdings, and items based on the bib record's CATKEY using the following strategy:

- Instance HRID uses the CATKEY with an `a` prefix
- Holding HRID uses the CATKEY with an `ah` prefix, the catkey, and then an underscore with a counter, i.e. `ah34578_1`, with counter being incremented for each holding record associated with the Instance.
- Item HRID uses the CATKEY with an `ai` prefix, the catkey, and then an underscore with a counter, i.e. `ai34578_1` with the counter being incremented for each item record associated with the Holding record.

~~NOTE: This PR has been rebased on the `t82-refactor-for-tsv` branch from PR #82~~

TODO:
- [x] Increase test coverage, especially the holding's and item's `_add_hrid` functions that are specific to Stanford.

### DIFF
--- a/plugins/folio/helpers.py
+++ b/plugins/folio/helpers.py
@@ -220,7 +220,7 @@ def setup_data_logging(transformer):
 
 
 def _apply_processing_tsv(tsv_path, airflow, column_transforms):
-    df = pd.read_csv(tsv_path, sep="\t")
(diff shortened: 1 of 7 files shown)


jgreben (approved):
LG2M

Issue #87, opened 2022-04-21
Error Parsing TSV


jermnelson:
```python
File pandas/_libs/parsers.pyx:1960, in pandas._libs.parsers.raise_parser_error()

ParserError: Error tokenizing data. C error: Expected 2 fields in line 3, saw 3
```
when parsing this example tsv file: 

```tsv
BARCODE	TECHSTAFF
36105010267412  	RECON/i:sv; atf:SAL 05/00 batch; i:reclass hbd20161025
36105045003279  	"{""TECHSTAFF"":[{""A"":3,""D"":""&#124;z added, i:nr 10/13/2020"",""I"":"""	"a"",""X"":0}]}"
36105038776741  	RECON/i:sv
36105005709295  	RECON/i:sv/i:rmr; tf: sal 1/95, rmr
36105005719088  	RECON/i:sv/i:rmr; tf: sal 1/95, rmr; MIDSPINE.suppl.
```

The problem is in the second line there is a tab between the `"""` and the "a".

jgreben:
This should be fixed now in the latest data dump.

PR #88, opened 2022-04-21
Added xfail test for bug report

jermnelson:
Inspired by this [blog](https://blog.ganssle.io/articles/2021/11/pytest-xfail.html) post on using pytest xfail as part of the bug report for #87. Putting this PR in draft because likely the fix is best done in the upstream Symphony exports.


jermnelson:
Fixed by upstream process.

PR #108, opened 2022-06-29
Split srs marc and process concurrently

jgreben:
Seems like this should kick off several marc srs batch post jobs at once...

Folio/postgres should be able to handle several concurrent threads, especially if we spin up multiple replicas of mod-source-record-storage. Maybe we can get a 5-hour job to finish in an hour or so....

@jermnelson let's pair on this if this approach is completely wrong...!

### DIFF
--- a/dags/srs_ingestion.py
+++ b/dags/srs_ingestion.py
@@ -57,17 +57,14 @@ def add_marc_to_srs():
         """
         split.bylinecount(Variable.get("NUMBER_OF_SRS_FILES", 10000))
 
-        for srs_file in srs_dir:
-            if srs_file.endswith('.ext'):
-                logger.info(f"Starting ingestion of {srs_file}")
-                post_marc_to_srs(
-                    dag_run=context.get("dag_run"),
-                    library_config=sul_config,
-                    srs_file=srs_file,
-                    MAX_ENTITIES=Variable.get("MAX_SRS_ENTITIES", 500),
-                )
-            else:
-                continue
+        for srs_file in Path(srs_dir).glob("*.ext"):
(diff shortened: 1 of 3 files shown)


jermnelson:
@jgreben, this makes senses. I think this approach should work.

dags/srs_ingestion.py
```diff
@@ -46,20 +49,31 @@ def ingestion_marc():
...
+        """
+        split.bylinecount(Variable.get("NUMBER_OF_SRS_FILES", 10000))
 
-        post_marc_to_srs(
-            dag_run=context.get("dag_run"),
-            library_config=sul_config,
-            srs_file=srs_filename,
-            MAX_ENTITIES=Variable.get("MAX_SRS_ENTITIES", 500),
-        )
+        for srs_file in srs_dir:
```
jermnelson, line 60: I'm not sure if this `for` loop actually does iterate through files in the `srs_dir` directory as one would expect. The `os.path.dirname` returns the directory path as a string. Instead, if you created a srs_dir Path object, you could use the glob method and eliminate the `if` statement. i.e.

```python
for srs_file in Path(srs_dir).glob("*.ext"):
     logger.info(f"Starting ingestion of {srs_file}")
     post_marc_to_srs(
         dag_run=context.get("dag_run"),
         library_config=sul_config,
         srs_file=srs_file,
         MAX_ENTITIES=Variable.get("MAX_SRS_ENTITIES", 500),
    )
```

jermnelson (changes requested):

dags/srs_ingestion.py
jgreben: Thanks!

jermnelson (approved):

PR #111, opened 2022-07-06
Supports dropping and restoring mod_inventory indices for SRS ingestion

### LINKED ISSUE #110: Investigating dropping Inventory Indexes before SRS ingestion
Index Data shared a Perl script that they use to drop the inventory indices for Instances, Holdings, and Items before loading SRS MARC records into FOLIO. Add these steps to the `add_marc_to_srs` DAG.

jermnelson:
Fixes #110 

Using the `PostgresOperator`, adds tasks to the `add_marc_to_srs` DAG:
Screen Shot 2022-07-06 at 4 07 49 PM

The `PostgresOperator` operators require a new  Postgres Connection with an id of `postgres_folio` that specifies the Postgres `host`, `username`, and `password`.

TODO:
- [ ] Finish testing from server (not locally because Postgres VM is only accessible from the Airflow server)
- [ ] Add tests (either by refactoring or by testing `add_marc_to_srs` DAG directly)

### DIFF
--- a/dags/bib_records.py
+++ b/dags/bib_records.py
@@ -320,7 +320,7 @@ with DAG(
         task_id="ingest-srs-records",
         trigger_dag_id="add_marc_to_srs",
         conf={
-            "srs_filename": "folio_srs_instances_{{ dag_run.run_id }}_bibs-transformer.json"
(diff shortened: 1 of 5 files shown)


jermnelson:
Closed in favor of #120

PR #126, opened 2022-07-29
Fix all the warnings that appear in the UI

jgreben:
- Remove broken warnings for Aeon to Lobby DAGS even after the variables are set
- Suppress robots.txt warning


jermnelson (approved):
Pending lint fixes

PR #132, opened 2022-08-12
New DAG for auto loading of bib records and tsv files

### LINKED ISSUE #131: Automate loading of BIB Record Loads
The current `symphony_marc_import` DAG's first task is a [FileSensor]() that monitors a directory for incoming MARC files (files ending in *.*rc) and the associated TSV files to trigger the next tasks in the DAG. Our current process is very manual, requiring us to copy the files into the mounted directory. 

Instead, we want to point to the directory and each DAG run will be initiated based on matching MARC and TSV files until the directory is empty.

jermnelson:
Fixes #131

### DIFF
--- a/dags/bib_records.py
+++ b/dags/bib_records.py
@@ -69,7 +69,7 @@ with DAG(
     start_date=datetime(2022, 1, 3),
     catchup=False,
     tags=["bib_import"],
-    max_active_runs=3,
+    max_active_runs=Variable.get("IMPORT_MAX_RUNS", 3),
 ) as dag:
 
     dag.doc_md = dedent(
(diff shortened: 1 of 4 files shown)


dags/bib_records.py
```diff
@@ -58,67 +62,36 @@
...
-
 with DAG(
     "symphony_marc_import",
     default_args=default_args,
-    schedule_interval=timedelta(minutes=15),
+    schedule_interval=None,
     start_date=datetime(2022, 1, 3),
     catchup=False,
     tags=["bib_import"],
+    max_active_runs=3,
```
shelleydoljack, line 72: Let's make this a parameter we can easily adjust instead of hardcoded.

shelleydoljack (approved):

dags/bib_records.py
jermnelson: Changed in latest commit, new `IMPORT_MAX_RUNS`

PR #140, opened 2022-09-12
DAG for processing preceding/succeeding titles

jgreben:
Fixes https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/142

### DIFF
--- a/plugins/folio/helpers.py
+++ b/plugins/folio/helpers.py
@@ -251,6 +251,41 @@ def _save_error_record_ids(**kwargs):
             error_file.write("\n")
 
 
+def put_to_okapi(**kwargs):
+    endpoint = kwargs.get("endpoint")
+    jwt = kwargs["token"]
+    records = kwargs["records"]
+    payload_key = kwargs["payload_key"]
+
+    tenant = "sul"
+    okapi_url = Variable.get("OKAPI_URL")
+
+    okapi_instance_url = f"{okapi_url}{endpoint}"
+
+    headers = {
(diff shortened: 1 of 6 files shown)


plugins/folio/helpers.py
```diff
@@ -286,6 +321,8 @@ def post_to_okapi(**kwargs) -> bool:
         logger.error(new_record_result.text)
         _save_error_record_ids(error_code=new_record_result.status_code, **kwargs)
 
+    return new_record_result.json
```
jgreben, line 324: For the test to pass I had to change this from `new_record_result.json()` so I am wondering if this is the right thing to return or not?

jermnelson: Hmm, this returns the method signature and not the JSON dictionary. I'm seeing some test failures like 

```python
    def __getattr__(self, name):
        if name in {'_mock_methods', '_mock_unsafe'}:
            raise AttributeError(name)
        elif self._mock_methods is not None:
            if name not in self._mock_methods or name in _all_magics:
>               raise AttributeError("Mock object has no attribute %r" % name)
E               AttributeError: Mock object has no attribute 'json'

/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/mock.py:634: AttributeError
```
Is this similar to what you're seeing?

jgreben: I was seeing `TypeError: 'dict' object is not callable` but using the lambda function as you suggested fixed it.

PR #154, opened 2022-09-20
Adds dashes in between date elements for catalogedDate.

### LINKED ISSUE #150: date cataloged should be the format YYYY-MM-DD
For readability and consistency with Lane's data, we should insert dashes between the timestamp elements when writing to catalgoedDate field.

shelleydoljack:
Closes #150

### DIFF
--- a/plugins/folio/instances.py
+++ b/plugins/folio/instances.py
@@ -19,7 +19,8 @@ def _adjust_records(bibs_transformer: BibsTransformer, tsv_dates: str):
             ckey = record["hrid"].removeprefix("a")
             matched_row = dates_df.loc[dates_df["CATKEY"] == ckey]
             if matched_row["CATALOGED_DATE"].values[0] != "0":
-                record["catalogedDate"] = matched_row["CATALOGED_DATE"].values[0]
+                date_cat = matched_row["CATALOGED_DATE"].values[0]
+                record["catalogedDate"] = date_cat[:4] + '-' + date_cat[4:6] + '-' + date_cat[-2:]
             records.append(record)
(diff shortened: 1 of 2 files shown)


plugins/folio/instances.py
```diff
@@ -19,7 +19,8 @@ def _adjust_records(bibs_transformer: BibsTransformer, tsv_dates: str):
             ckey = record["hrid"].removeprefix("a")
             matched_row = dates_df.loc[dates_df["CATKEY"] == ckey]
             if matched_row["CATALOGED_DATE"].values[0] != "0":
-                record["catalogedDate"] = matched_row["CATALOGED_DATE"].values[0]
+                date_cat = matched_row["CATALOGED_DATE"].values[0]
+                record["catalogedDate"] = date_cat[:4] + '-' + date_cat[4:6] + '-' + date_cat[-2:]
```
jermnelson, line 23: Another option would be to use python datetime's `strptime` and `strftime` methods:

```python
from datetime import datetime
created_date = datetime.strptime("20131014", "%Y%m%d")
record['catalogedDate'] = created_date.strftime("%Y-%m-%d")
```
The advantage of using these methods is that an exception will be raised if the parsed date isn't in YYYYmmdd format and the formatted date string would always have the year, month, and day in the correct locations.

jermnelson (approved):
Approved with a suggestion to use `datetime` methods instead of string slices for date format.

plugins/folio/instances.py
shelleydoljack: Ahh, good suggestion.

PR #130, opened 2022-08-10
Uses MARC field 596 for LIBRARY column in Electronic Holdings TSV

jermnelson:
Fixes #129

PR #160, opened 2022-09-23
Simplifies Electronic Holdings and uses 596 fields for LIBRARY column

jermnelson:
Per @ahafele [comment](https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/72#issuecomment-1248214285), removes 856 URI and notes from Electronic Holdings, supersedes PR  [130](https://github.com/sul-dlss/libsys-airflow/pull/130) by including code to generate the LIBRARY column from the 596 fields in the MARC record.

- ~~investigate adding these electronic holdings to the base tsv file and remove more code later in the DAG~~

### DIFF
--- a/plugins/folio/helpers/marc.py
+++ b/plugins/folio/helpers/marc.py
@@ -8,9 +8,6 @@ import pymarc
 
 from pathlib import Path
 
-from airflow.models import Variable
-
-from folioclient import FolioClient
 from folio_migration_tools.migration_tasks.batch_poster import BatchPoster
 
 
@@ -33,18 +30,16 @@ def _add_electronic_holdings(field856: pymarc.Field) -> bool:
     return True
 
 
-def _extract_856s(catkey: str, fields: list, relationships: dict) -> list:
+def _extract_856s(**kwargs) -> list:
+    catkey = kwargs["catkey"]
+    fields = kwargs["fields"]
+    library = kwargs["library"]
+
     properties_names = [
         "CATKEY",
         "HOMELOCATION",
         "LIBRARY",
-        "LINK_TEXT",
         "MAT_SPEC",
-        "PUBLIC_NOTE",
-        "RELATIONSHIP",
-        "URI",
-        "VENDOR_CODE",
-        "NOTE",
     ]
     output = []
     for field856 in fields:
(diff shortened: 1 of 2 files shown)


plugins/folio/helpers/marc.py
```diff
@@ -8,80 +8,81 @@
...
-            row["NOTE"] = "|".join(all_x_subfields)
-        if field856.indicator2 in relationships:
-            row["RELATIONSHIP"] = relationships[field856.indicator2]
-        else:
-            row["RELATIONSHIP"] = relationships["_"]
         output.append(row)
     return output
 
 
+def _get_library(fields596: list) -> str:
```
jgreben, line 70: Would a switch statement work better here?

jgreben (approved):
Just a suggestion to use a case/switch block instead of if/elif/else. Tale it or leave it.

plugins/folio/helpers/marc.py
```diff
@@ -8,80 +8,81 @@
...
 from folio_migration_tools.migration_tasks.batch_poster import BatchPoster
 
 
 logger = logging.getLogger(__name__)
 
 
 full_text_check = re.compile(
-    r"(table of contents|abstract|description|sample text)", re.IGNORECASE
+    r"(table of contents)|(abstract)|(description)|(sample text)", re.IGNORECASE
 )
```
shelleydoljack, line 19: Regex difference between python and ruby? double-checked here: https://github.com/sul-dlss/searchworks_traject_indexer/blob/6e47b81cc894b61bc01bdcdbc80e363546bb9f50/lib/traject/config/sirsi_config.rb#L800

plugins/folio/helpers/marc.py
```diff
@@ -8,80 +8,81 @@
...
+    if field856.indicator2 in ["0", "1"]:
         subfield_z = field856.get_subfields("z")
         subfield_3 = field856.get_subfields("3")
         subfield_all = " ".join(subfield_z + subfield_3)
-        if full_text_check.match(subfield_all):
+        if full_text_check.search(subfield_all):
             return False
-    return True
+        return True
+    return False
```
shelleydoljack, line 32: @ahafele Do we care about 856 2nd indicator blank? Should that meet the criteria for an e-holdings?

plugins/folio/helpers/marc.py
jermnelson: Yeah, originally I tried to use a match/case syntax (the python "switch" statement) here but the complicating factor is that each "if" statement is evaluating a condition (that the string number in the subfield_a) and the case statement does not allow evaluation statements.

plugins/folio/helpers/marc.py
jermnelson: Thanks for pointing this out @shelleydoljack! The problem wasn't in the original regex (I just restored the original one that matches Seachworks) but that I was using `match` instead of `search` on line 29.

plugins/folio/helpers/marc.py
ahafele: I do not think it should meet the criteria for an e-holdings. I think we'll end up with alot of bad holdings that way.

PR #155, opened 2022-09-21
Refactor item notes to fix errors

jermnelson:
CLOSES #151 
CLOSES #152

TODO:
- [x] Fix remaining tests and increase code coverage

PR #163, opened 2022-09-26
Adds library and catalog format to item type for HOOVER and HV-ARCHIVE.

### LINKED ISSUE #158: Concatenate HOOVER item types
For the items tsv files, when the LIBRARY column is HOOVER or HV-ARCHIVE, then concatenate (with space) the values of ITEM_TYPE FORMAT LIBRARY columns and save the resulting value into the ITEM_TYPE column. This is for material type mapping. Depends on https://github.com/sul-dlss/folio_migration/issues/18 
e.g. ITEM_TYPE = STKS MARC HOOVER

shelleydoljack:
Closes #158 

Merging this will affect PR #155


jermnelson (approved):
Looks good

PR #164, opened 2022-09-27
Removes dates tsv file from Symphony mount.

### LINKED ISSUE #157: dates.tsv files should be moved from the symphony mount after processing
After the auto_bib_loads and symphony_marc_import DAGs run, the *.dates.tsv files are still in the mount on the symphony server.

shelleydoljack:
Closes #157

### DIFF
--- a/plugins/folio/instances.py
+++ b/plugins/folio/instances.py
@@ -2,6 +2,7 @@ import json
 import logging
 import pandas as pd
 from datetime import datetime
+from pathlib import Path
 
 from folio_migration_tools.migration_tasks.bibs_transformer import BibsTransformer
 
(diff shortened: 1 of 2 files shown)


jermnelson (approved):

jgreben (approved):

jgreben (changes requested):
Maybe just write a test so coveralls is happy..

jermnelson:
Approving but could add the following to test at https://github.com/sul-dlss/libsys-airflow/blob/578a34b6eadc6cc14133e7cfd2468cb34038a84c/plugins/tests/test_instances.py#L50, 
```assert not tsv_dates_file.exists()``` to keep the current coverage.

shelleydoljack:
I thought about writing a test too, but I don't think testing the `.unlink()` function is necessary. I can't stick `assert not tsv_dates_file.exists()` into `test_adjust_records` because `_adjust_records` doesn't do the unlink, it is the `run_bibs_transformer` method that does. I tried looking at re-writing `test_run_bibs_transformer` but it is more involved to do that for testing an python function. 🤷

jgreben:
Can you just move the unlink() to the _adjust_records function then? Otherwise, what do you want to do about the coverage? Usually the inability to test something means something probably needs to be refactored.

jgreben (approved):

PR #167, opened 2022-09-28
Fixes item type modification.

### LINKED ISSUE #165: Only modify item type for HOOVER or HV-ARCHIVE
https://github.com/sul-dlss/libsys-airflow/blob/578a34b6eadc6cc14133e7cfd2468cb34038a84c/plugins/folio/helpers.py#L390-L393

This updates all the item types if the DataFrame (the entire tsv file) has any row with LIBRARY=HOOVER or HV-ARCHIVE. This should either iterate over the rows in the DataFrame (although I'm reading that's not a good way to go about this) or pass a specific row to this method to modify the item type (but again, that would be iterating over the entire DataFrame). There's probably another, better solution.

shelleydoljack:
Close #165

### DIFF
--- a/plugins/folio/helpers.py
+++ b/plugins/folio/helpers.py
@@ -387,8 +387,9 @@ def _apply_transforms(df, column_transforms):
 
 
 def _modify_item_type(df, libraries):
-    if df["LIBRARY"].isin(libraries).any():
-        df["ITEM_TYPE"] = df["ITEM_TYPE"] + " " + df["FORMAT"] + " " + df["LIBRARY"]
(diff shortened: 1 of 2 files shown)


jermnelson (approved):
Approving, looking at the coverage drop https://coveralls.io/builds/52847290/source?filename=plugins%2Ffolio%2Fhelpers.py, none of this code is implicated.

jgreben (changes requested):
Maybe check coveralls and see what still needs testing.

jgreben (approved):

PR #168, opened 2022-09-29
Checks for BIB Record Loads without Item Notes

jermnelson:
Otherwise raises exceptions when trying to load non-existent notes.

### DIFF
--- a/plugins/folio/helpers.py
+++ b/plugins/folio/helpers.py
@@ -439,19 +439,20 @@ def _processes_tsv(tsv_base: str, tsv_notes: list, airflow, column_transforms, l
         logging.info(f"Merged {len(note_df)} notes into items tsv")
         tsv_note_path.unlink()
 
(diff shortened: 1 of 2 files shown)


jgreben (approved):
The PR description says something about raising an exception for non-existent notes. Did you intend to include that (I don't see where it does that if you did ¯\_(ツ)_/¯ ). If you do want to do that then maybe also a test for that. Otherwise fine as it is.

shelleydoljack (approved):

jermnelson:
> The PR description says something about raising an exception for non-existent notes. Did you intend to include that (I don't see where it does that if you did ¯_(ツ)_/¯ ). If you do want to do that then maybe also a test for that. Otherwise fine as it is.

Sorry, should have been more clear; without these checks, missing tsv notes files raise a Pathlib exception trying to create an file object on a nil variable. This PR does a check first before trying to create a `pathlib.Path` object.

PR #194, opened 2022-11-08
Adds 956 fields to create electronic holdings

jermnelson:
Fixes https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/236

### DIFF
--- a/plugins/folio/helpers/marc.py
+++ b/plugins/folio/helpers/marc.py
@@ -21,10 +21,10 @@ vendor_id_re = re.compile(r"\w{2,2}4")
 sdr_sul_re = re.compile(r"https*:\/\/purl.stanford.edu")
 
 
-def _add_electronic_holdings(field856: pymarc.Field) -> bool:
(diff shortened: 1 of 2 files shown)


shelleydoljack (approved):

plugins/folio/helpers/marc.py
```diff
@@ -21,18 +21,18 @@
...
+        subfield_3 = field.get_subfields("3")
         subfield_all = " ".join(subfield_z + subfield_3)
         if full_text_check.search(subfield_all):
             return False
         return True
     return False
 
 
-def _extract_856s(**kwargs) -> list:
+def _extract_fields(**kwargs) -> list:
```
jgreben, line 35: Maybe consider renaming this to something like `extract_e_holdings_fields` since it only applies to them and not all marc fields?

jermnelson: Done, updated in latest commit.

jgreben (approved):

PR #200, opened 2022-11-29
Fixes errors due to processing and removes forked dependency

jermnelson:
Fixes some slight errors that surfaced during the latest complete migration run. If missing `preceding_succeeding_titles.json`  file for DAG run and adds iteration_id for capturing error file.  PR also restores upstream `folio_migration_tools` with the recent [release](https://github.com/FOLIO-FSE/folio_migration_tools/releases/tag/1_6_3) that now generates files like `preceding_succeeding_titles.json` so as not to depend on logging.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,5 @@ COPY requirements.txt .
 
 RUN pip install -r requirements.txt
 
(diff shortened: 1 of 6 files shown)


jermnelson:
The `folio_migration_tools` also includes a fix for 880 fields.

shelleydoljack (approved):

PR #202, opened 2022-12-06
Circ Rules Tester App

### LINKED ISSUE #201: Create circ_rules_tester DAG
Based on https://github.com/sul-dlss/libsys-airflow/tree/circ-report branch, create/update DAG for `circ_rules_tester`

jermnelson:
Fixes #201 

New Circ Rules Tester App allows users to either enter in UUIDs for a single scenario test or upload a csv file with a list of scenarios and run tests.

Screen Shot 2022-12-07 at 5 02 57 PM

TODO:
- [x] Finish batch DAG with supporting code


shelleydoljack (approved):

jgreben:
I guess I have questions about this..and I apologize for not paying enough attention or being more involved in the discussions about the requirements for this (I guess I was busy/distracted by the various things breaking folio)..

- where do the end-users find the uuids to enter? Is there not any friendly names that can be populated in lists for them to pick from?
- how does folio_migration fit into circ_rules_tester?
- Does it take a long time to run and then produce a report to be fetched? If no, why a DAG? 
- Should something like actually this be in libsys-webforms, or do we plan on using airflow as the place to serve up folio "webforms" instead of the rails app...? Ultimately/ideally there should probably only be one place where people go for supplemental reports and apps, unless this is of a different kind that needs to be in airflow for some reason?

jermnelson:
> I guess I have questions about this..and I apologize for not paying enough attention or being more involved in the discussions about the requirements for this (I guess I was busy/distracted by the various things breaking folio)..

No worries; totally understand  :)

### * where do the end-users find the uuids to enter? Is there not any friendly names that can be populated in lists for them to pick from?

Not entirely sure; I think they find them from Folio or from spreadsheets. One of the requests from demoing the app was having drop-down lists.

### * how does folio_migration fit into circ_rules_tester?
The only relationship between `folio_migration` and `circ_rules_tester` is FOLIO. The original demos were created in Jupyter notebook from existing Python code.

 ### * Does it take a long time to run and then produce a report to be fetched? If no, why a DAG?
No, the reports typically run quickly and I don't think a DAG is necessarily the best approach.

 ### * Should something like actually this be in libsys-webforms, or do we plan on using airflow as the place to serve up folio "webforms" instead of the rails app...? Ultimately/ideally there should probably only be one place where people go for supplemental reports and apps, unless this is of a different kind that needs to be in airflow for some reason?

No reason why this couldn't be in libsys-webforms; I definitely agree that having a single place for these types of supplemental apps is definitely preferable. We may want to think about ways we could integrate the Rails front-end with backend Python code (both the circ_rules_tester and the SpineOmatic app I've been investigating were originally developed in Python). Being able to leverage existing projects instead of re-implementing everything in Ruby would definitely be a time saver.

jgreben (approved):

Issue #206, opened 2023-01-11
New strategy for generating Item UUIDs


jermnelson:
Currently we use the HRID of the Item along with Okapi URL and record type to generate UUID for the Item.  https://github.com/sul-dlss/libsys-airflow/blob/2215e9caf38bae25fed58c30c021fe8c1dc74b71/plugins/folio/helpers/folio_ids.py#L222

We want to change to the following:

1. Use the barcode of the Item first if it exists
2. If barcode is empty string, uses columns CATKEY, CALL_SEQ, COPY into a single string without a delimiter

PR #209, opened 2023-01-20
Use item uuid generated from folio_migration_tools and migration mapping

jgreben:
Uses https://github.com/sul-dlss/folio_migration/pull/44 to allow the migration tools to create the unique item UUID by using a concatenation of the ckey+barcode. In the case of items with barcodes the combination will be unique. IN the case of on-order items without barcodess the ckey alone will always be unique because there are no duplicate ckeys in the generating tsv files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,9 +78,9 @@ The `optimistic_locking_management` DAG requires a Postgres Airflow
 database being used by Okapi.
 
 ## Testing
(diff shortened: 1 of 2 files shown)


jgreben:
Fixes #206

jermnelson (approved):

PR #212, opened 2023-01-27
Adds function to create new holdings and boundwith post-process file

jgreben:
Also refactors mock of xcom push/pull to be called from mocks file.

### DIFF
--- a/dags/bib_records.py
+++ b/dags/bib_records.py
@@ -37,6 +37,7 @@ from plugins.folio.holdings import (
     post_folio_holding_records,
     run_holdings_tranformer,
     run_mhld_holdings_transformer,
+    boundwith_holdings,
 )
 
 from plugins.folio.login import folio_login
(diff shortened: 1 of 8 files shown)


dags/bib_records.py
```diff
@@ -179,7 +180,8 @@
         convert_tsv_to_folio_holdings >> convert_tsv_to_folio_items
         convert_tsv_to_folio_items >> finish_conversion
 
-    with TaskGroup(group_id="mhlds-electronic-holdings") as mhlds_electronic_holdings:
+    # with TaskGroup(group_id="mhlds-electronic-holdings") as mhlds_electronic_holdings:
```
jermnelson, line 183: Remove this line?

dags/preceding_succeding_titles.py
```diff
@@ -29,7 +29,7 @@ def post_to_folio(*args, **kwargs):
     params = context.get("params")
     iteration_id = params["iteration_id"]
     task_instance = kwargs["task_instance"]
-    jwt = task_instance.xcom_pull(task_ids="folio_login")
+    jwt = task_instance.xcom_pull(task_ids="folio_login", key="")
```
jermnelson, line 32: I think this will error when running in the DAG because a blank string key won't exist for the task instance. Below is a suggestion on how to fix the mock so the test will pass.

plugins/tests/mocks.py
```diff
@@ -100,14 +100,35 @@ def mock_file_system(tmp_path, mock_dag_run):
...
+# Mock xcoms
+def mock_xcom_push(*args, **kwargs):
+    key = kwargs["key"]
+    value = kwargs["value"]
+    messages[key] = value
+
+
+def mock_xcom_pull(*args, **kwargs):
+    task_id = kwargs["task_ids"]
+    key = kwargs["key"]
```
jermnelson, line 116: You can change this to use a `get` method so it won't raise a `KeyError` when running the preceding_succeding_titles test; i.e.:

```python
key = kwargs.get("key")
```

jermnelson (changes requested):

PR #225, opened 2023-02-28
Circ Rules Tester Loan Fix and Mocks Refactoring

jermnelson:
The circ rules tester now ensures that policies and other Okapi results have limits parameter set at a high level to grab all of the records for these rule endpoints. 

Also, updates test to use the XCOM mocks and refactors the mocks message handling to include a task_id for the `MockInstance`.

### DIFF
--- a/plugins/folio/circ_rules.py
+++ b/plugins/folio/circ_rules.py
@@ -235,6 +235,7 @@ def generate_urls(**kwargs):
             "location_id": instance.xcom_pull(
                 task_ids="setup-circ-rules", key=f"location_id{row_count}"
             ),
+            "limit": 2_000  # Should be plenty
         }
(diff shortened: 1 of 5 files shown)


plugins/folio/circ_rules.py
```diff
@@ -235,6 +235,7 @@ def generate_urls(**kwargs):
             "location_id": instance.xcom_pull(
                 task_ids="setup-circ-rules", key=f"location_id{row_count}"
             ),
+            "limit": 2_000  # Should be plenty
```
shelleydoljack, line 238: I sure hope so! (Please, let's not have more than 2,000 circ rules...)

shelleydoljack (approved):

PR #229, opened 2023-03-06
Assigns Instance statusId based on Cataloged Date

jermnelson:
Fixes https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/308


shelleydoljack (approved):
👍

PR #230, opened 2023-03-08
Removal of previous identifiers from Instances and Holdings

jermnelson:
Fixes https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/299


shelleydoljack (approved):
I'm glad we're getting rid of this; they're not needed, especially the formerIds for holdings records.

PR #234, opened 2023-03-10
New SRS Remediation DAG

### LINKED ISSUE #227: SRS Audit-Remediation DAG
Similar to the current audit-remediation DAG, create a DAG that would run after a complete SRS migration to check if both BIB and MHLD MARC SRS records are loaded into FOLIO and attempt to add these records if there is an error.

jermnelson:
Fixes #227 

Using the `audit-remediation.db` database, this new DAG is designed to be run after all of the SRS ingestion DAGs are completed. Checks and attempts to add any missing SRS MHLD or Bib records to FOLIO.

Screen Shot 2023-03-10 at 10 07 38 AM

### DIFF
--- a/dags/srs_remediation.py
+++ b/dags/srs_remediation.py
@@ -1,0 +1,151 @@
+import json
+import logging
+import pathlib
+import tarfile
+import uuid
+
+import requests
+
+from datetime import datetime, timedelta
+
+from airflow import DAG
+from airflow.operators.empty import EmptyOperator
+from airflow.models import Variable
+from airflow.operators.python import PythonOperator
+
+from folioclient import FolioClient
+
+logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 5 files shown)


shelleydoljack (approved):
👍 
I was trying to figure out if the srs_total_sql was actually doing a call to SRS to see how many bib or holdings records there are, because if so, we have to take into account Lane's records (which we aren't loading). But it looks like we are querying our own database of how many SRS records we sent to to be posted to SRS?

jermnelson:
> 👍 I was trying to figure out if the srs_total_sql was actually doing a call to SRS to see how many bib or holdings records there are, because if so, we have to take into account Lane's records (which we aren't loading). But it looks like we are querying our own database of how many SRS records we sent to to be posted to SRS?

Yes, this is just querying the local sqlite3 audit database and not the SRS.

PR #252, opened 2023-03-28
Meta-programming to support MHLD 852 Map

jermnelson:
This PR overrides the upstream `folio_migration_tools` package's `fix_853_bug_in_rules` method that was stripping out the 852 rule, subfield 'z' for setting a HoldingsStatement note that fixed a bug where the mapping was a creating a `permanentLocationId` property in the holdings statement. 

To fix now re-introduced bug, created a decorator function  `_wrap_additional_mapping`, that wraps the RuleMapperBase.perform_additional_mapping to move the `permanentLocationId` property to the top-level in the Holdings record. 

**NOTE** other changes are the result of running the Black code formatter and not intentional changes 🥲

### DIFF
--- a/plugins/folio/holdings.py
+++ b/plugins/folio/holdings.py
@@ -69,18 +69,30 @@ def _alt_get_legacy_ids(*args):
 
 def _wrap_additional_mapping(func):
     """
-    Decorator that moves the `permanentLocationId` property from the
-    holdingsStatements property to a top-level property in the Holdings
-    Record.
+    Decorator that moves the top-level properties from the holdingsStatements
+    in the Holdings Record.
     """
+    top_level_props = {
+        "callNumberTypeId",
+        "permanentLocationId",
+        "callNumber",
+        "callNumberPrefix",
+        "shelvingTitle",
+        "callNumberSuffix",
+        "copyNumber"
+    }
+
     def wrapper(*args, **kwargs):
         holdings_record = args[1]
-        if "permanentLocationId" not in holdings_record:
-            for statement in holdings_record.get("holdingsStatements", []):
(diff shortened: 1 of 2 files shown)


plugins/folio/holdings.py
```diff
@@ -66,6 +67,24 @@ def _alt_get_legacy_ids(*args):
...
+        if "permanentLocationId" not in holdings_record:
+            for statement in holdings_record.get("holdingsStatements", []):
+                if "permanentLocationId" in statement:
+                    holdings_record["permanentLocationId"] = statement.pop(
+                        "permanentLocationId"
+                    )
+        func(*args, **kwargs)
+    return wrapper
+
+
```
shelleydoljack, line 87: Just double-checking, do the other properties listed as targets in the mapping for 852 (callNumberTypeId, permanentLocationId, callNumber, callNumberPrefix, shelvingTitle, callNumberSuffix, copyNumber) also need to be moved out of holdingsStatement? (was checking the [holdingsrecord.json](https://github.com/folio-org/mod-inventory-storage/blob/master/ramls/holdingsrecord.json)).

shelleydoljack (commented):
👍 merge if the other targets don't need moving too.

plugins/folio/holdings.py
jermnelson: Good question! I didn't see these other targets in the holdingsStatements but let me run a test to find out.

jermnelson: @shelleydoljack - I updated the `_wrap_additional_mapping` function to move all of the top-level properties to the holdings record if you want to review again.

shelleydoljack (approved):

PR #259, opened 2023-03-30
Removes ON-ORDER Items from TSV

### LINKED ISSUE #253: do not create folio item records for on order items
For all the tsv files coming from Symphony (the ckey_range.tsv and ON-ORDER.*.tsv files), if the CURRENTLOCATION or HOMELOCATION column has the value ON-ORDER, create just a holdings record and no item record in folio.

jermnelson:
Fixes #253 Also does some DRY refactoring for when post-processing Items.

### DIFF
--- a/plugins/folio/items.py
+++ b/plugins/folio/items.py
@@ -15,6 +15,34 @@ from plugins.folio.helpers import post_to_okapi, setup_data_logging
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


shelleydoljack (approved):
This PR reminds me that we need to discuss what happens to any item xinfo notes that were on the item records with ON-ORDER home/current location.

PR #275, opened 2023-04-04
Updating development instructions in README

lwrubel:
Updates a few issues I ran into when setting up libsys-airflow for development. There are likely other sections of the README that would benefit from an update (see Deployment) but I did not test and update those.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,16 +13,14 @@ Based on the documentation, [Running Airflow in Docker](https://airflow.apache.o
 > docker daemon, we recommend at least 5GB.
 
 1. Clone repository `git clone https://github.com/sul-dlss/libsys-airflow.git`
-2. Run `git submodule init` and then `git submodule update` to populated submodule.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,9 +87,10 @@ The `optimistic_locking_management` DAG requires a Postgres Airflow
 database being used by Okapi.
 
 ## Testing
-- Activate python virtual environment: `source virtual-env/bin/activate`
-- Install the FOLIO-FSE tools: `pip install folio_migration_tools`
-- Install Airflow: `pip install apache-airflow apache-airflow-providers-postgres`
+1. If you haven't already, create a python virtual environment in the home directory: `python3 -m venv virtual-env`
+2. Activate the virtual environment: `source virtual-env/bin/activate`
+3. Install dependencies:  `pip install -r requirements.txt`
+4. Install Airflow: `pip install apache-airflow==2.5.3 apache-airflow-providers-postgres==5.4.0`
```
lwrubel, line 93: I'm not a big fan of having versions in the README. These aren't pinned/locked anywhere else at the moment, so in order to keep this in sync with the current `Dockerfile`, putting those versions here.

jermnelson (approved):

PR #282, opened 2023-04-06
Skeleton DAG for vendor marc bib loads

### LINKED ISSUE #247: Create a Scaffold for a Daily Airflow DAG for retrieving dynamic list of MARC BIB records from various vendors
Create an Airflow DAG that runs daily and queries the Dataloader Management App and possibly FOLIO [Organizations](https://dev.folio.org/reference/api/#mod-organizations-storage)  for a dynamic list of vendors that includes connection information (FTP, SFTP, username, passwords) and Data Import Profile for MARC BIB records along with specific pre- and post- processing tasks to be preformed on these records. 

This ticket should trigger secondary DAGs for each vendor load. 

Specific tasks to be accomplished will be ticketed separately.

lwrubel:
Resolves #247 to set up a scaffold for a DAG to handle daily vendor marc loads.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -34,6 +34,7 @@ jobs:
         flake8 --ignore=E225,E501,F401,F811,W503 dags/ plugins/
     - name: Test with pytest
       run: |
+        airflow db init
         coverage run -m pytest
         coverage lcov
         coveralls --service=github
--- a/dags/vendor_marc_bib_load.py
+++ b/dags/vendor_marc_bib_load.py
@@ -1,0 +1,41 @@
+from datetime import datetime, timedelta
+import logging
+
+from airflow import DAG
+
+from airflow.operators.empty import EmptyOperator
+
+logger = logging.getLogger(__name__)
+
+
+default_args = {
+    "owner": "folio",
+    "depends_on_past": False,
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "retry_delay": timedelta(minutes=5),
+}
+
+
+with DAG(
+    dag_id="vendor_marc_bib_load",
+    default_args=default_args,
+    schedule="0 0 * * *",
+    catchup=False,
+    start_date=datetime(2023, 1, 1)
+) as dag:
+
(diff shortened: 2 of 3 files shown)


dags/vendor_marc_bib_load.py
```diff
@@ -0,0 +1,41 @@
...
+
+from airflow import DAG
+
+from airflow.operators.empty import EmptyOperator
+
+logger = logging.getLogger(__name__)
+
+
+default_args = {
+    "owner": "folio",
```
lwrubel, line 12: Is this the correct owner, or should it be "libsys"?

dags/vendor_marc_bib_load.py
```diff
@@ -0,0 +1,41 @@
...
+    "email_on_retry": False,
+    "retries": 0,
+    "retry_delay": timedelta(minutes=5),
+}
+
+
+with DAG(
+    dag_id="vendor_marc_bib_load",
+    default_args=default_args,
+    schedule="0 0 * * *",
```
lwrubel, line 24: Set to midnight since I don't think we've determined what time exactly it should run.

.github/workflows/python-app.yml
```diff
@@ -34,6 +34,7 @@ jobs:
         flake8 --ignore=E225,E501,F401,F811,W503 dags/ plugins/
     - name: Test with pytest
       run: |
+        airflow db init
```
lwrubel, line 37: Allows us to test loading dags.

dags/vendor_marc_bib_load.py
jermnelson: Good question; we've been using `folio` as the owner but I don't think it matters from this [article](https://stackoverflow.com/questions/68085083/what-is-the-owner-property-in-airflow-used-for).

jermnelson (approved):
Approving and merging. The drop in coverage is not because of this PR but because we have started testing DAGs directly and the existing DAGs are now included. Will create a follow-up ticket to ignore these DAGs through coverage [config file](https://coverage.readthedocs.io/en/latest/source.html#execution).

PR #283, opened 2023-04-06
Add healthcheck app.

justinlittman:
refs https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/353

image


jermnelson (approved):
Looks good! From the quality of this PR, you wouldn't know that it has been five years since you worked in Python. :)

PR #284, opened 2023-04-06
Use poetry to manage python dependencies

### LINKED ISSUE #276: Add support for dependency management using Poetry
Add [Poetry](https://python-poetry.org/) for dependency management.

mjgiarlo:
Fixes #276

This commit changes how python dependencies are managed in libsys-airflow, switching from `pip` to `poetry`. This particularly affects the DAGs and plugins we load into the docker-run Airflow instance. To load these poetry-managed dependencies into the running Airflow instance, we integrate `poetry build` and `pip` via the wheel packaging standard.

As part of this work, we move the directories that need the poetry-managed dependencies into a new directory, `libsys_airflow`, as that is a requirement of `poetry build`, so a large number of renames (moves) are included in this commit. This directory is now where all of the python work is handled.


README.md
```diff
@@ -30,8 +33,8 @@ Based on the documentation, [Running Airflow in Docker](https://airflow.apache.o
 ### Prerequisites
 1. Install `pip3` with `apt install python3-pip`
 1. Install python virtual enviroments: `apt install python3.8-venv`
-1. Create the virtual envirnment in the home directory: `python3 -m venv virtual-env`
-1. Install docker-compose in the virtual environment: `source virtual-env/bin/activate && pip3 install docker-compose`
+1. Install dependencies per `Dependency Management and Packaging` above
+1. Install docker-compose in the poetry virtual environment: `poetry shell && pip3 install docker-compose`
```
mjgiarlo, line 37: One of the changes I'll want help testing is around the venv changes here. It's not clear to me how venv was being used before (single purpose? multiple?) and whether the changes I've made here in the README are broken or unnecessary.

pyproject.toml
```diff
@@ -0,0 +1,34 @@
...
+pytest-mock = "^3.7.0"
+
+[tool.poetry.dev-dependencies]
+apache-airflow = "2.5.3"
+apache-airflow-providers-postgres = "5.4.0"
+coveralls = "~3.3.1"
+coverage = "~6.5.0"
+flake8 = "^6.0.0"
+folioclient = "^0.43.0"
+folio-uuid = "^0.2.8"
```
mjgiarlo, line 30: Manage all dependencies (prod and dev/test alike) in one place instead of `pip install`s sprinkled all over and commingling dependencies.

docker-compose.yaml
```diff
@@ -69,16 +69,17 @@ x-airflow-common:
...
-    - ${AIRFLOW_PROJ_DIR:-.}/logs:/opt/airflow/logs
-    - ${AIRFLOW_PROJ_DIR:-.}/plugins:/opt/airflow/plugins
-    - ${AIRFLOW_PROJ_DIR:-.}/symphony:/opt/airflow/symphony
+    - ${AIRFLOW_PROJ_DIR:-.}/libsys_airflow/dags:/opt/airflow/dags
+    - ${AIRFLOW_PROJ_DIR:-.}/libsys_airflow/logs:/opt/airflow/logs
+    - ${AIRFLOW_PROJ_DIR:-.}/libsys_airflow/plugins:/opt/airflow/plugins
+    - ${AIRFLOW_PROJ_DIR:-.}/libsys_airflow/symphony:/opt/airflow/symphony
     - ${AIRFLOW_PROJ_DIR:-.}/migration:/opt/airflow/migration
-    - ${AIRFLOW_PROJ_DIR:-.}/circ:/opt/airflow/circ
+    - ${AIRFLOW_PROJ_DIR:-.}/libsys_airflow/circ:/opt/airflow/circ
```
mjgiarlo, line 82: Move the working directories into the new python/poetry dir. Tidies up the root and makes clear where the python stuff's happening.

docker-compose.yaml
```diff
@@ -69,16 +69,17 @@ x-airflow-common:
     # See https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/logging-monitoring/check-health.html#scheduler-health-check-server
     # yamllint enable rule:line-length
     AIRFLOW__SCHEDULER__ENABLE_HEALTH_CHECK: 'true'
+    SQLALCHEMY_SILENCE_UBER_WARNING: 1
```
mjgiarlo, line 72: These muddy the docker-compose logs.

config/deploy.rb
```diff
@@ -56,7 +56,7 @@
   desc 'install airflow dependencies'
   task :install do
     on roles(:app) do
-      execute "cd #{release_path} && source #{fetch(:venv)} && pip3 install -r requirements.txt"
+      execute "cd #{release_path} && pip3 install -r requirements.txt && poetry build --format=wheel --no-interaction --no-ansi && pip3 install dist/*.whl"
```
mjgiarlo, line 59: I don't think we need the venv here?

README.md
```diff
@@ -48,12 +51,9 @@ cap airflow:webserver      # restart webserver
 
 ### Do the first time you bring up Libsys-Airflow:
 1. Log into the server, `ksu` and install `apt install python3.8-venv libpq-dev`
-1. Exit `ksu` and install the pyton virtual environment under the libsys home directory:
-   1. `cd && mkdir virtual-env`
-   1. `python3 -m venv /home/libsys/virtual-env/`
 1. In your local environment do `cap ${stage} deploy`
 1. Follow the instructions for [shared_configs/libsys-airflow](https://github.com/sul-dlss/shared_configs/tree/libsys-airflow#readme)
-1. On the server `source /home/libsys/virtual-env && pip install docker-compose==1.29.2`
+1. On the server `poetry shell && pip3 install docker-compose==1.29.2`
```
mjgiarlo, line 56: I notice we use `pip` in some places and `pip3` in others. I'm not sure if we should care about consistency, or if we genuinely need this to vary in different contexts. 🤷🏻‍♂️

.github/workflows/python-app.yml
```diff
@@ -20,21 +20,18 @@ jobs:
...
       run: |
-        flake8 --ignore=E225,E501,F401,F811,W503 dags/ plugins/
+        cd libsys_airflow
+        poetry run flake8 --ignore=E225,E501,F401,F811,W503 dags/ plugins/
     - name: Test with pytest
       run: |
-        airflow db init
-        coverage run -m pytest
-        coverage lcov
-        coveralls --service=github
```
mjgiarlo, line 40: Poetry handles all the deps now, so we no longer need to install new ones in CI.

pyproject.toml
lwrubel: I think these are also development dependencies:

```
pytest
pytest-mock
pydantic
```

mjgiarlo: @lwrubel I found that those needed to be prod dependencies, or airflow was unhappy. If a dependency is declared in the prod dependencies above, does it need to be explicitly listed as a dev dep?

Dockerfile
```diff
@@ -9,6 +9,9 @@ ENV SLUGIFY_USES_TEXT_UNIDECODE "yes"
 
 USER airflow
 
-COPY requirements.txt .
+COPY requirements.txt pyproject.toml poetry.lock .
+COPY libsys_airflow ./libsys_airflow
 
 RUN pip install -r requirements.txt
+RUN poetry build --format=wheel --no-interaction --no-ansi
+RUN pip install dist/*.whl
```
lwrubel, line 17: I'm curious about using `poetry build` and then pip to install the wheels, rather than a regular `poetry install`. Is that a recommended practice for poetry in building a Docker image?

README.md
```diff
@@ -87,10 +87,9 @@ The `optimistic_locking_management` DAG requires a Postgres Airflow
 database being used by Okapi.
 
 ## Testing
-1. If you haven't already, create a python virtual environment in the home directory: `python3 -m venv virtual-env`
-2. Activate the virtual environment: `source virtual-env/bin/activate`
-3. Install dependencies:  `pip install -r requirements.txt`
-4. Install Airflow: `pip install apache-airflow==2.5.3 apache-airflow-providers-postgres==5.4.0`
+1. Install dependencies per `Dependency Management and Packaging` above
+1. Change into the poetry build dir: `cd libsys_airflow`
+1. Drop into the poetry virtual environment: `poetry shell`
```
lwrubel, line 92: The first time I ran through these instructions, I had to do `poetry install --with dev` to get the airflow dependency. Then I could run `poetry shell` and `pytest` successfully. 

I did a `docker build` once, to build the image and bring Airflow up. That worked. 

Now, starting again with a new poetry environment (I deactivated and removed the initial one, to start fresh), when doing `poetry build --format=wheel` I get:
```
'/opt/airflow/logs/scheduler/2023-04-07' is not in the subpath of '/Users/laurawrubel/sul-dlss/libsys-airflow' OR one path is relative and the other is absolute.`
```

I'm not sure if I messed things up with this order of steps, but happy to pair and show what I'm seeing.

Dockerfile
mjgiarlo: @lwrubel I naïvely started doing `poetry install`, which is typically how I've used `poetry`. When I did that, Airflow seemingly did not "see" the dependencies when it started up. As far as I could tell, this two-step process is key to installing the `poetry`-managed dependencies into the Airflow environment (which relies on `pip`). Let me confirm that, though, and I'll drop another reply.

README.md
mjgiarlo: Hmm. I had been seeing that before too until I added the `exclude` and `include` directories in `pyproject.toml`, and then that error went away for me. Maybe that wasn't the right fix, though. Thanks for testing!

Dockerfile
mjgiarlo: Using `poetry install` in docker-land, causes Airflow's plugins to be unhappy:

```
airflow-webserver_1  | [2023-04-07 15:11:06,985] {plugins_manager.py:266} ERROR - Failed to import plugin /opt/airflow/plugins/folio/
helpers/marc.py                                                                                                                      
airflow-webserver_1  | Traceback (most recent call last):                                                                            
airflow-webserver_1  |   File "/home/airflow/.local/lib/python3.10/site-packages/airflow/plugins_manager.py", line 258, in load_plugi
ns_from_plugin_directory                                                                                                             
airflow-webserver_1  |     loader.exec_module(mod)                                                                                   
airflow-webserver_1  |   File "<frozen importlib._bootstrap_external>", line 883, in exec_module                                     
airflow-webserver_1  |   File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed                                
airflow-webserver_1  |   File "/opt/airflow/plugins/folio/helpers/marc.py", line 10, in <module>                                     
airflow-webserver_1  |     import pymarc                                                                                             
airflow-webserver_1  | ModuleNotFoundError: No module named 'pymarc'           
```

config/deploy.rb
jermnelson: We were using a Python 3.10 virtual environment because the VM's version of Python is 3.8 and is incompatible with the `folio_migration_tools`. This may not be necessary now that we're using poetry so further testing is needed.

README.md
jermnelson: When in the virtual environment, `pip` and `pip3` are identical. When not using a virtual environment, pip3 is usually installed through Linux package manager and manages the system's Python 3 dependencies. When using `poetry shell` that activates a virtual environment, the `pip3` should be the local version and not the system version.

README.md
mjgiarlo: @lwrubel I tossed a change at this branch if you want to re-pull and try again. I'm trying to verify here but my local python dev env is all mucked up because ... I don't currently known enough to unmuck it up. Happy to pair!

README.md
mjgiarlo: @jermnelson OK, thank you! Happy to change anything here, since my local python dev env is janky.

config/deploy.rb
mjgiarlo: Oh, then that is likely still broken (and probably related to some of the local python woes I'm having because ubuntu). I'll revert this change.

pyproject.toml
jermnelson: FWIW  pydantic is `folio_migration_tools` dependency as well as development dependency that we are using for testing.

docker-compose.yaml
jermnelson: Cool! Nice addition.

config/deploy.rb
mjgiarlo: Reverted this change.

config/deploy.rb
```diff
@@ -27,7 +27,7 @@
 desc 'Change release directories ownership'
 task :fix_permissions do
   on roles(:app) do
-    within release_path do
+    within releases_path do
```
mjgiarlo, line 30: This was a bug. (Unrelated to poetry but discovered when trying to re-deploy.)

config/deploy.rb
```diff
@@ -130,7 +130,7 @@
     on roles(:app) do
       execute "docker image prune -f"
       execute "cd #{release_path} && releases=($(ls -tr ../.)) && cd ../${releases[0]} && source #{fetch(:venv)} && docker-compose stop"
-      execute "docker rm $(docker ps --filter status=exited -q)"
+      execute "[[ $(docker ps -aq) ]] && docker ps -aq | xargs docker stop | xargs docker rm || echo 'no containers to stop'"
```
mjgiarlo, line 133: A problem I discovered when re-deploying.

lwrubel (approved):

pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[tool.poetry]
+name = "libsys-airflow"
+version = "0.1.0"
+description = "Airflow DAGS for migrating and managing ILS data into FOLIO along with other LibSys workflows"
+authors = ["Jeremy Nelson <jermnelson@gmail.com>", "Joshua Greben <jgreben@stanford.edu>", "Shelley Doljack", "Laura Wrubel", "Michael J. Giarlo <mjgiarlo@stanford.edu>"]
```
jermnelson, line 5: Could you change this to use my Stanford email `jpnelson@stanford.edu` instead of my personal gmail account?

libsys_airflow/airflow.cfg
jermnelson, line 1: Does Airflow pick-up this configuration with it being moved to the `libsys_airflow` directory?

README.md
```diff
@@ -4,7 +4,9 @@
 Airflow DAGS for libsys processes and migrating ILS data into FOLIO
 
 ## Dependency Management and Packaging
-Run `pip install -r requirements.txt` to install the dependencies.
+To install the dependencies, run:
+* `pip install -r requirements.txt`
+* `poetry install`
```
jermnelson, line 9: I tried following these directions on the sul-libsys-airflow-prod and I ran into a conflict with the system's Python 3.8 version and poetry. The first line, `pip installed -r requirements` worked fine but then when I tried to the second command `poetry install`, I saw this error: 

```
 poetry install
/usr/lib/python3/dist-packages/requests/__init__.py:89: RequestsDependencyWarning: urllib3 (1.26.15) or chardet (3.0.4) doesn't match a supported version!
  warnings.warn("urllib3 ({}) or chardet ({}) doesn't match a supported "
The currently activated Python version 3.8.10 is not supported by the project (^3.10).
Trying to find and use a compatible version. 
```
Python 3.10.10 is currently installed on this VM in order to use the latest version of docker-compose so maybe we don't even need to install Poetry and the project's dependencies.

jermnelson: This could very much be operator error on my part if the intention is not to install poetry on the server but for local development.

jermnelson (changes requested):
One minor request for changing email with a few non-blocking questions.

README.md
mjgiarlo: @jermnelson Yeah, those instructions are only for local use. The steps that need to be run on the server are in the Dockerfile and in deploy.rb.

libsys_airflow/airflow.cfg
mjgiarlo: @jermnelson Good question, because I'm not sure. How can we tell if `warn_deployment_exposure` has the right value in the running instance?

pyproject.toml
mjgiarlo: Yep!

mjgiarlo: Done.

mjgiarlo:
@jermnelson 💬 
> One minor request for changing email with a few non-blocking questions.

Other than not knowing whether airflow.cfg is picked up, this should be ready to go. Thanks for the review!

libsys_airflow/airflow.cfg
jermnelson: I logged into the running container and ran the following [command](https://airflow.apache.org/docs/apache-airflow/stable/howto/set-config.html) 

```bash
airflow@580cf45ee73f:/opt/airflow$ airflow config get-value webserver warn_deployment_exposure
True
```
 So I don't think it is being set correctly. Maybe copy this file to the `/opt/airflow/`?

jermnelson (approved):

Issue #287, opened 2023-04-07
Use environment variables to configure Airflow


mjgiarlo:
Currently, every time Airflow is re-deployed, a user must go into the Airflow UI and copy over configuration values from shared_configs to get Airflow running properly. Instead we want to inject these values via environment variables.

- [x] Store values in vault: `aeon_url`, `folio_user`, `folio_password`, `okapi_url` (documented in `shared_configs`: https://github.com/sul-dlss/shared_configs/blob/folio-airflow/README.md#variables-for-folio-development)
  - [x] dev
  - [x] prod
  - [x] stage (once the box is created)
- [x] Set environment variables based on vault lookups: `AIRFLOW_VAR_AEON_URL`, `AIRFLOW_VAR_FOLIO_USER`, `AIRFLOW_VAR_FOLIO_PASSWORD`, `AIRFLOW_VAR_OKAPI_URL`
  - [x] dev (except Aeon URL which is not currently in `shared_configs`)
  - [x] prod
  - [x] stage (once the box is created)
- [x] Make necessary changes to `Dockerfile` and/or `docker-compose.yaml` to pass the above environment variables through to containerland (see https://github.com/sul-dlss/libsys-airflow/pull/288)
- [x] Change python code in `libsys_airflow/dags` and `libsys_airflow/plugins` to access environment variables. E.g., instead of `Variable.get("OKAPI_URL")`, use `Variable.get("okapi_url")`, which Airflow will look up in the env at `AIRFLOW_VAR_OKAPI_URL`. (See [Airflow env var documentation](https://airflow.apache.org/docs/apache-airflow/stable/howto/variable.html#storing-variables-in-environment-variables))

A good chunk of this work is already done for dev, and is sitting in this draft PR: https://github.com/sul-dlss/libsys-airflow/pull/288 Feel free to run with that or dump it.

lwrubel:
@jermnelson Should `AEON_KEY` also be included in this work?

Will you be able to populate vault with the values for prod and stage? (Or point me to where I would find those, and I can add them, I think.) That seems necessary for when this would eventually get merged and deployed.

jmartin-sul:
@jermnelson says `AEON_KEY` yes, `MAX_SRS_ENTITIES` probably not.

jermnelson:
Sure, keeping this ticket open until stage is ready makes sense.

mjgiarlo:
Stage is ready and is configured to use vault for env vars. I believe we can close this now.

PR #288, opened 2023-04-07
Inject Airflow environment variables in docker-compose config

mjgiarlo:
Connects to #287

This commit allows the running airflow services to be configured dynamically via environment variables.


Dockerfile
```diff
@@ -9,7 +9,7 @@ ENV SLUGIFY_USES_TEXT_UNIDECODE "yes"
 
 USER airflow
 
-COPY requirements.txt pyproject.toml poetry.lock .
+COPY requirements.txt pyproject.toml poetry.lock ./
```
lwrubel, line 12: When using COPY with more than one source file, Docker expects the destination to be a directory and end with a `/`

lwrubel:
Puppet PR to accompany this: https://github.com/sul-dlss/puppet/pull/9411/files (will take it out of draft once this PR is reviewed to confirm it's the right set of variables).

All of the variables are in vault for dev. Prod and stage will need the new ones.

jermnelson (changes requested):
Looks good overall. Just a request to update the README at https://github.com/sul-dlss/libsys-airflow#running-locally-with-docker to add these variables to the local `.env` file

lwrubel:
Thanks, that's a good suggestion. Instructions added.

jermnelson (approved):

PR #295, opened 2023-04-11
Move test dependencies

### LINKED ISSUE #274: Breakout requirements for production Docker Images
Currently we're including dev dependencies in our Docker environment when building with our default image from  https://github.com/sul-dlss/libsys-airflow/blob/631242e90fd5936baa6d2e33af5a2b8f1f74f1eb/Dockerfile#L13-L15.

edsu:
Move test dependencies into `test` poetry group, and modify the Docker build to not install test dependencies. It looks like building a wheel with poetry and pip installing the wheel only includes the main dependencies and not dependencies from other groups, like test.

Since the `tests` were part of the `libsys_airflow` module the airflow app complained about missing `pytest` once it was no longer being installed. The tests were moved out to the top level directory, which required quite a few files needed their imports updated to reference `libsys_airflow`.

Fixes #274


jmartin-sul (approved):
LGTM, but refraining from merging since review was requested from jeremy

jermnelson (approved):

PR #299, opened 2023-04-13
remove the specification of Ruby version from Gemfile

jmartin-sul:
Ruby is only used for Capistrano deployments, most if not all of us likely default to Ruby 3.x on our laptops, and the 2.7 branch of Ruby is EOL around now, see https://www.ruby-lang.org/en/downloads/branches/

> Ruby 2.7
>
> status: security maintenance
> release date: 2019-12-25
> normal maintenance until: 2022-04-01
> EOL: 2023-03-31 (expected)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,6 @@
 
 source 'https://rubygems.org'
 
-ruby '2.7.1'
-
 group :deployment do
   gem 'capistrano'
   gem 'capistrano-bundler'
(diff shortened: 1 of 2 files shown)


jmartin-sul:
fwiw: i did a little looking around, and i noticed that some libsys projects specify ruby version in Gemfile this way, some leave it unspecified, as is the case after this PR.  spot checking infra and access projects, i didn't see any that specified ruby version in this way.

PR #301, opened 2023-04-13
Switch to managed DB for prod.

justinlittman:


ndushay (approved):
approving with eyeballs but not much knowledge

PR #303, opened 2023-04-13
Example Gobi Interface DAG

jermnelson:
This draft PR is an example of a DAG for a specific Vendor/Interface based on Gobi vendor Yankee-SUL Interface:

Screenshot 2023-04-13 at 12 47 56 PM

Comments on XCOM are inline


libsys_airflow/dags/gobi.py
```diff
@@ -0,0 +1,115 @@
...
+
+    delay_invoice_file = EmptyOperator(
+        task_id="delay-invoice-processing",
+    )
+
+    extract_zipfiles = PythonOperator(
+        task_id="extract-zipfiles",
+        python_callable=zip_extraction,
+        op_kwargs={
+            "zipfile": "{{ ti.xcom_pull('check-count-file', key='zipfile') }}"
```
jermnelson, line 36: There are two main methods of passing XCOM messages, one by through the task template and the second through a kwargs variable in the called Python function.

libsys_airflow/plugins/vendors/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+
+from airflow.operators.python import get_current_context
+
+logger = logging.getLogger(__name__)
+
+def backup_retrieved_files(*args, **kwargs):
+    """
+    Reference ticket https://github.com/sul-dlss/libsys-airflow/issues/265
+    """
+    task_instance = kwargs["task_instance"]
```
jermnelson, line 12: To access XCOM messages from upstream tasks, the `task_instance` is one of the passed in kwargs from the PythonOperator.

libsys_airflow/plugins/vendors/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+    backup_files = kwargs["backup_files"]
+    zip_task_id = kwargs["zip_task_id"]
+    zip_files = task_instance.xcom_pull(task_ids=zip_task_id)
```
jermnelson, line 15: One advantage of pull XCOM messages directly in the function is that the message can be more than just a string, in this example it is a python list (unlike the first XCOM method, all python objects are passed through as strings)

libsys_airflow/plugins/vendors/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+
+def check_retrieve_files(*args, **kwargs):
+    context = get_current_context()
+    interface_info = context.get('params')
+    task_instance = kwargs["task_instance"]
+
+    # Connects
+    logger.info(f"Connects to Vendor {interface_info}")
+    interface_path = "/opt/airflow/vendor_123/2023-04-13" 
+    task_instance.xcom_push(key="marc-brief-orders", value=f"{interface_path}/sample.mrc")
```
jermnelson, line 28: Push to task's XCOM with a specific key/value for retrieval by downstream tasks.

libsys_airflow/plugins/vendors/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+def zip_extraction(*args, **kwargs):
+    """
+    Reference ticket https://github.com/sul-dlss/libsys-airflow/issues/262
+    """
+    # Extract
+    zipfile = kwargs["zipfile"]
+    path = pathlib.Path(zipfile)
+    logger.info(f"Zip extraction for {zipfile}")
+    # Returning value from a task's function is put into the default XCOM
+    return [f"{path.parent}/sample-01.mrc", f"{path.parent}/sample-02.mrc"]
```
jermnelson, line 61: If a task's function returns a value, it is automatically added as an available XCOM message with the `return_value` as the key.

edsu:
Do you think it would be ok to write a `mock_dag_run` test for this to make sure it runs and then merge it into main so we can start filling it in?

libsys_airflow/dags/gobi.py
```diff
@@ -0,0 +1,115 @@
...
+    rename_vendor_files,
+    zip_extraction
+)
+
+with DAG(
+    "gobi",
+    schedule_interval=None,
+    start_date=datetime(2023, 4, 11),
+    catchup=False,
+) as dag:
```
aaron-collier, line 20: If we want to use the airflow rest API in a flask interface to get dag(s) and dagrun status, it would be useful to use tags here so that we can query by tag:  https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html#operation/get_dags

I think at least 2 tags would be useful: `vendor` (literal), `org_id_from_folio` to query for the specific vendor, and `data type` (bib/invoice/etc).

This would give us data that can be used to filter properly for the flask UI.

jermnelson:
> This is super helpful @jermnelson. Do you think it would be ok to write a `mock_dag_run` test for this to make sure it runs and then merge it into main so we can start filling it in?

Hi @edsu, I'll work on getting test completed and get this PR out of draft status.

jermnelson:
Closing this PR as other work has superseded the need for this example.

PR #305, opened 2023-04-14
Remove extra template and update README

justinlittman:
Edit by @lwrubel: It seems that the Variables are necessary for importing DAGs and including viewing plugin apps. Locally, these are added via a `.env`.


lwrubel:
It wouldn't hurt for someone else to test this locally.

ndushay (changes requested):
This PR removes the healthcheck ... was that unintentional?  Is a rebase needed?

lwrubel:
Yes, it is a duplicate, leftover from before the move of the directory. See `libsys-airflow/libsys_airflow/plugins/folio/templates/healthcheck/index.html`

jmartin-sul (approved):

lwrubel:
@ndushay are there still any concerns? May I merge this?

PR #309, opened 2023-04-17
Update dependencies with Github Action

### LINKED ISSUE #257: Add dependency updates and deployment
Should leverage current Github [CI-actions](https://github.com/sul-dlss/libsys-airflow/blob/main/.github/workflows/python-app.yml)

edsu:
This Github Action will run every Monday at 3am UTC and use Poetry to update the "main" dependencies. If there are any changes it will use the github command line tool to create a "dependency updates" pull request.

Only dependencies in the "main" section of `pyproject.toml` are updated because `apache-airflow` is installed by the tests, and we are using an Airflow Docker image for deployment. If we let Poetry update `apache-airflow` it can potentially install a newer version on top of the one that we are using in the Docker container.

You can see a PR it created (as a test) here: https://github.com/sul-dlss/libsys-airflow/pull/315

Fixes #257

### DIFF
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -1,0 +1,37 @@
+name: Update Dependencies
+
+on:
+  schedule:
+    - cron: '0 3 * * 1'
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+jobs:
+  update:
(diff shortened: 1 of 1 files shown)


justinlittman (approved):

ndushay:
I wonder if it's worthwhile explaining why no airflow is loaded in a comment in the script.

PR #311, opened 2023-04-17
Adds support for creating/updating Connections.

### LINKED ISSUE #298: Airflow Connection Management for Vendor Interfaces
As required by #261, for each Vendor's Interface, we need to have a corresponding Airflow Connection that contains the host and login credentials. With the FOLIO Interface record being the "source-of-truth", we need to:

- Check if an Airflow Connection already exists for the vendor Interface 
- Create a new Airflow Connection if doesn't
- Update an existing Airflow Connection if the Vendor Interface changes

See https://stackoverflow.com/questions/51863881/is-there-a-way-to-create-modify-connections-through-airflow-api

justinlittman:
closes #298


libsys_airflow/plugins/airflow/connections.py
```diff
@@ -0,0 +1,20 @@
+from airflow.models import Connection
+from sqlalchemy.orm.session import Session
+from airflow.utils.session import NEW_SESSION, provide_session
+
+
+@provide_session
+def find_or_create_conn(conn_type, host, login, pwd, session: Session = NEW_SESSION):
+    conn_id = f'{conn_type}-{host}'
```
aaron-collier, line 8: I wonder if it would make sense to use the interface UUID from Folio as the connection ID in case of race conditions between possible tasks for the same vendor/host but a different interface? It may not matter as the host/login/pwd more than likely don't change but I assume it's possible.

justinlittman: The reason that I generated a connection id is because it is human-readable. Since the connections are listed in the UI, I thought a human-readable connection id would be helpful for troubleshooting.

aaron-collier: Good point.

README.md
```diff
@@ -109,10 +109,10 @@ Install the test database (sqlite):
 `airflow db init`
 
 Then, to run the test suite, use [pytest](https://docs.pytest.org/).
-`pytest`
+`poetry run pytest`
```
jermnelson, line 112: If you're already in the `poetry shell` (in step 2) then `pytest` would be sufficient. Maybe change this to `poetry run pytest` or `pytest` after running poetry shell.

README.md
```diff
@@ -109,10 +109,10 @@ Install the test database (sqlite):
...
 
 To see stdout or stderr add the `-rP` flag:
-`pytest -rP`
+`poetry run pytest -rP`
```
jermnelson, line 115: See comment above

jermnelson (changes requested):
Looks good; suggest changing the README directions.

README.md
justinlittman: Gotcha. Removed.

README.md
justinlittman: Removed.

PR #320, opened 2023-04-18
Add apache-airflow-client to Python dependencies

### LINKED ISSUE #318: Add airflow python client to dependencies
Install the airflow python client and ensure it is in the dependency list: https://github.com/apache/airflow-client-python

mjgiarlo:
Fixes #318

Version 2.5.1 of the client, the latest version released as of this commit, should be compatible with the Airflow 2.5.3 API judging by the 2.5.2 and 2.5.3 release notes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,8 @@ pymarc = "^4.1.3"
 folio-migration-tools = "1.7.8"
 requests = "^2.28.2"
 pyyaml = "^6.0"
+# this should ideally be kept in sync with the airflow version but client releases run behind server releases
+apache-airflow-client = "2.5.1"
 
 [tool.poetry.group.test.dependencies]
 coveralls = "~3.3.1"
--- a/poetry.lock
+++ b/poetry.lock
@@ -360,6 +360,22 @@ yandex = ["apache-airflow-providers-yandex"]
 zendesk = ["apache-airflow-providers-zendesk"]
 
 [[package]]
+name = "apache-airflow-client"
+version = "2.5.1"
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -29,6 +29,7 @@ pytest-mock = "^3.7.0"
 # the airflow version here should be kept in sync with the docker image
 apache-airflow = "2.5.3"
 apache-airflow-providers-postgres = "5.4.0"
+apache-airflow-client = "2.5.1" # this should ideally be kept in sync with the airflow version but client releases are lagging
```
aaron-collier, line 32: I think this needs to be in the main dependencies as we'll use it for interacting with airflow in the flask UI.

aaron-collier (changes requested):

pyproject.toml
edsu: Kinda wondering if we should talk about how to install and run Airflow using Poetry in the Docker container?

mjgiarlo: I'm not saying it's not possible, but IIRC the Airflow documentation explicitly supports installation by `pip` and not by `poetry`. 🤷🏻‍♂️

mjgiarlo:
@aaron-collier @edsu Is this PR good to go now or should we make changes?

aaron-collier:
@mjgiarlo apologies, I've been meaning to get back to this. Yes, i think we still need it as task code will be required to instantiate dag runs, likely through the client, but if we find we can do everything in models, we can always back this out if we find it unused.

aaron-collier (approved):

PR #328, opened 2023-04-24
[EXAMPLE] Adding database queries and dev docker-compose.yml var

lwrubel:
Ephemeral branch with some example code (from a tutorial) for getting data from the existing Airflow database.


jmartin-sul:
no longer need this example, and it's still findable as an example even if not merged

PR #329, opened 2023-04-24
SQLAlchemy Models for Datamanagement App Plugin

jermnelson:
Taking a slightly different approach, this branch uses a new database ( running as a separate "database" within the Postgres Airflow Postgres environment) with ORM models for Organizations, Interfaces, JobRuns, Files, and JobStatus for managing the state of data loading. 

Would like to be able to use the [Flask Appbuilder](https://flask-appbuilder.readthedocs.io/en/latest/quickhowto.html) to build not only the Models but the CRUD views as well but was getting an error trying to run under Airflow. Instead, just uses the plain SQLAlchemy Models but could also look to use the [Flask SQLAlchmey](https://flask-sqlalchemy.palletsprojects.com/) for better integration with the plugin.

### DIFF
--- a/libsys_airflow/plugins/data_loading_management/data_loading.py
+++ b/libsys_airflow/plugins/data_loading_management/data_loading.py
@@ -1,0 +1,24 @@
+from airflow.models import Variable
+
+from flask_appbuilder import expose, BaseView as AppBuilderBaseView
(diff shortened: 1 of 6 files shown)


aaron-collier:
Closing in favor of the alembic approach merged this morning.

PR #331, opened 2023-04-25
[HOLD | EXAMPLE] Initial stubs for Vendor Load Plugin

aaron-collier:
This is a draft branch to provide an example of extending the airflow UI with a flask based plugin.


jmartin-sul:
no longer need this example, and it's still findable as an example even if not merged

Issue #332, opened 2023-04-25
Establish Vendor Management plugin in airflow menu and initial vendor list


aaron-collier:
Using https://github.com/sul-dlss/libsys-airflow/pull/331 as an example of extending the airflow dashboard menu to include a flask based plugin:

- [x] Add a "Vendor Management" menu to airflow
- [x] Add initial vendor list (screenshot example below, queried from folio organizations) that includes vendor name, code, id, and acquisitions unit. 
- [ ] Vendor list is paged and searchable by name and code

This initial page is complete when the vendor list displays all organizations in folio, linking to configuration and history pages will be addressed in followup tickets. The status, last scheduled job, and buttons in the example below should _**not**_ be included.

<img width="1466" alt="Screenshot 2023-04-25 at 8 41 45 AM" src="https://user-images.githubusercontent.com/2294288/234331857-4af51d7c-9205-4382-a30d-360614fb9814.png">

lwrubel:
A few questions:
* The mockup shows the menu item as "Data Loading Management" but the text requests "Vendor Management". Which one should be used? I would use the same wording in naming the plugins directory. 
* Should the result of getting the organizations from FOLIO be written to the database (which is still in progress?). I think the answer is yes, but wasn't sure if that should also be included with this. 

I think it might be worth splitting the content of the Vendors page into a separate ticket.

lwrubel:
Based on conversation with Aaron, will use Vendors Management as the naming throughout and will break up into multiple PRs.

lwrubel:
As the ticket requests, I'm querying the live FOLIO for the needed organizations. But I think this is temporary, until the database and the vendors table is set up to serve as the source of organizations here? I am trying to figure out how much I should be testing the interaction with FOLIO.

PR #350, opened 2023-05-01
Vendor Load: Setting up list of vendors from FOLIO

lwrubel:
Further work on #332. This sets up data in the vendor page, but is pulling from live Folio until the database vendor table is available. 

Screenshot 2023-05-01 at 11 03 24 AM

### DIFF
--- a/libsys_airflow/plugins/vendor/vendors.py
+++ b/libsys_airflow/plugins/vendor/vendors.py
@@ -1,3 +1,5 @@
+from folioclient import FolioClient
+import logging
 import requests
 
 from airflow.models import Variable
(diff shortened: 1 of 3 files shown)


tests/apps/test_vendor_management_view.py
```diff
@@ -1,7 +1,35 @@
...
+def mock_get_folio_client(monkeypatch):
+    def mock_folio_client():
+        folio_client = MockFOLIOClient()
+        return folio_client
+
+    monkeypatch.setattr(VendorManagementView, "_folio_client", mock_folio_client)
+
+
+def test_vendor_management_view(mock_okapi_variable, mock_okapi_requests, mock_get_folio_client):
     vendor_management_app = VendorManagementView()
```
lwrubel, line 34: Not testing much at this point, since the live Folio query is temporary and will be replaced with a database query in the near term.

libsys_airflow/plugins/vendor/vendors.py
```diff
@@ -1,14 +1,45 @@
...
 
 from flask_appbuilder import expose, BaseView as AppBuilderBaseView
 
 
+logger = logging.getLogger(__name__)
+
+
 class VendorManagementView(AppBuilderBaseView):
     default_view = "vendors_index"
     route_base = "/vendor"
```
edsu, line 15: Would it be helpful to use the plural `/vendors` here since the view is of a list of vendors? I always struggle with this when naming routes...

edsu (approved):

libsys_airflow/plugins/vendor/vendors.py
lwrubel: That does make more sense to me, @edsu. @aaron-collier: the plugin being called `vendor` (in the `plugins/vendor` directory) is what led me to use vendor instead, matching that. Work for you if I switch the route base to `vendors`? Or is this specified somewhere?

aaron-collier: Yeah - I'd suggestion `/vendors` (I may have suggested this incorrectly earlier) as that matches how rails routes work as well.

aaron-collier (approved):

PR #351, opened 2023-05-01
Support FTP downloads.

### LINKED ISSUE #250: Airflow Task: use FTP to Download MARC Records
Create an airflow task for downloading vendor data using the [FTP Operator](https://airflow.apache.org/docs/apache-airflow-providers-ftp/stable/index.html) for a vendor interface.

Follow the pattern established in DLME for reusable task code: https://github.com/sul-dlss/dlme-airflow/blob/main/dlme_airflow/tasks/harvest.py

This task module should take the configuration JSON as a parameter and using the provided vendor_id and interface_id query the FOLIO okapi API for the interface credentials. This will likely require using https://github.com/sul-dlss/libsys-airflow/blob/main/libsys_airflow/plugins/airflow/connections.py to dynamically establish an airflow connection.

This task is complete when:
- [ ] a task module (i.e. .../libsys-airflow/libsys_airflow/tasks/fetcher.py) exists
- [ ] the task module returns a PythonOperator that can be included in a DAG
- [ ] the code called by the PythonOperator returns a list of downloaded files (this allows the results to be used through the XCOM return value of the task.
- [ ] the task uses the config JSON payload to establish a connection and retrieve all files from the vendor that match the filename regex provided
- [ ] task task code has sufficient test coverage

Notes: 
- For the initial iteration of this task, it is expected that any files that match the provided filename regex are downloaded.
- The files are saved to **{data_loader_path}/{vendor_id}/{interface_id}** where data_load_path comes from the env (this will be /dataloader/data on the server(s)), vendor_id and interface_id come from the config. If the directory does not exist, create it.
- It may be worthwhile, but not required, to look into adding a command line utility for running this task individually during development (https://github.com/sul-dlss/dlme-airflow/blob/main/bin/post-harvest)

justinlittman:
closes #250


libsys_airflow/plugins/vendor/marc_filter.py
```diff
@@ -3,22 +3,34 @@
 
 import pymarc
 
+from airflow.decorators import task
```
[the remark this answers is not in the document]
justinlittman: https://github.com/psf/black

libsys_airflow/plugins/airflow/connections.py
```diff
@@ -1,10 +1,26 @@
...
+    info = interface_info(interface_id=interface_id)
+    host = info['uri'].removeprefix('ftp://')
+    conn_id = find_or_create_conn('ftp', host, info['username'], info['password'])
+    logger.info(f"Connection id for Interface {interface_id}: {conn_id}")
+    return conn_id
+
 
 @provide_session
-def find_or_create_conn(conn_type, host, login, pwd, session: Session = NEW_SESSION):
+def find_or_create_conn(conn_type: str, host: str, login: str, pwd: str, session: Session = NEW_SESSION) -> str:
```
ndushay, line 23: why do we need explicit typing here?

libsys_airflow/plugins/vendor/marc_filter.py
```diff
@@ -3,22 +3,34 @@
...
+
 def filter_fields(marc_path: pathlib.Path, fields: list):
     """
     Filters specified fields from MARC records
     """
     logger.info(f"Filtering fields from {marc_path}")
     filtered_records = []
-    with marc_path.open('rb') as fo:
-        marc_reader = pymarc.MARCReader(fo, to_unicode=True, permissive=True, utf8_handling='replace')
+    with marc_path.open("rb") as fo:
```
ndushay, line 25: is this implication that we still don't have a linter for this repo?   Or that the linter is agnostic on quote style?

tests/vendor/test_download.py
```diff
@@ -0,0 +1,28 @@
...
+import logging
+
+import pymarc
+
+from libsys_airflow.plugins.vendor.download import download
+
+@pytest.fixture
+def hook(mocker):
+    mock_hook = mocker.patch('airflow.providers.ftp.hooks.ftp.FTPHook')
+    mock_hook.list_directory.return_value = ['3820230411.mrc', '3820230412.mrc', '3820230412.xxx']
```
ndushay, line 14: reading this test code, I'm assuming the code is checking for filenames, so it doesn't matter that there are no such fixture files.  Is this correct?

libsys_airflow/dags/vendor/test.py
```diff
@@ -0,0 +1,68 @@
...
+default_args = {
+    "owner": "folio",
+    "depends_on_past": False,
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "retry_delay": timedelta(minutes=5),
+}
+
+with DAG(
```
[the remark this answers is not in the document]
justinlittman: I went with "data_fetcher" since this matches the design doc.

ndushay (approved):
a few more "I'm new to python" questions but ow LGTM.

libsys_airflow/plugins/airflow/connections.py
justinlittman: No compelling reason.

libsys_airflow/plugins/vendor/marc_filter.py
justinlittman: This is beyond my knowledge of python linting / formatting.

tests/vendor/test_download.py
justinlittman: That is correct. This is testing that it ignores files that have already been downloaded or don't match the regex.

PR #346, opened 2023-04-30
Vendor Load: Establish pattern for model development and initial database setup

aaron-collier:
The goal of this PR is to setup an initial model that may be used as a pattern for the additional model tickets to follow this. The database migration/initializations here should be considered temporary until a proper migration strategy can be implemented.

Until #344 is merged, you will need to manually create the database in postgres with airflow as the owner.

PR #352, opened 2023-05-01
Vendor Load: Use Alembic for Vendor Load Database Migrations

### LINKED ISSUE #337: Investigate and document using flask_migrate or alembic for database initialization/migration
https://flask-migrate.readthedocs.io/en/latest/

Flask migrate looks like a useful tool to implement rails like database migrations for a flask app. Using https://github.com/sul-dlss/libsys-airflow/pull/346 as an example of models we will implement, investigate implement using flask_migrate to:

- [ ] initialize the database for the vendor management app
- [ ] establish the basic vendor model and create the table
- [ ] document in the README the flask/flask_migrate commands that developers will use to initialize the vendor management database for local development
- [ ] determine how (or if) we can automate flask_migrate based database migrations on deployment

Alternatively investigate using Alembic (https://alembic.sqlalchemy.org/en/latest/autogenerate.html) which I think flask-migrate is based on.

jermnelson:
Fixes #337 

Based on branch in  PR [346](https://github.com/sul-dlss/libsys-airflow/pull/346), this PR uses [Alembic](https://alembic.sqlalchemy.org/en/latest/) for managing database migration with a single migration script for creating the Vendor Model. This approach is very similar to how Rails handles migrations with ActiveRecord. 

A couple of decisions that merit discussion:
- After first testing with having the alembic migration directory in the same directory as the `libsys_airflow/plugins/vendor` directory, decided to move it back to the root directory so that migration scripts would not be copied into the Docker image when being built. Could move it back but I don't think this migration code is needed in Docker
- Using environmental variables for the Database credentials are documented in the README


README.md
```diff
@@ -109,6 +109,29 @@ The `optimistic_locking_management` DAG requires a Postgres Airflow
...
+
+### Vendor load plugin
+
+Using and developing the vendor load plug in requires it's own database. Ensure that the `vendor_loads` database exists in your local postgres and is owned by the airflow user.
+
+#### Database migrations
+Using [Alembic](https://alembic.sqlalchemy.org/en/latest/) to manage database migrations for the `vendor_loads` database, changes to the 
+Models are autogenerated in a migration file in the `vendor_loads_migration/versions` directory. The Alembic migration requires the following environmental variables to be set either through a local `.env` file or 
+injected when using docker-compose:
+- **DATABASE_USERNAME**
```
edsu, line 122: I wonder if the environment variables should indicate that they are for the Vendor Load App instead of being just generic? So VENDOR_LOAD_DB_USERNAME instead of DATABASE_USERNAME?

edsu (commented):
This looks good to me, and I'm glad you were able to get Alembic to manage migrations without requiring the dropping of data! 

I liked the idea of colocating the db configuration and migrations with the vendor load app itself. But I can live with them being separate. I'm confused, don't the migrations need to run in the deployed environments (e.g. prod) and thus need to be copied into the Docker container so that they can be run from in there?

I think one follow on ticket from this (if we don't have it already) is to get Capistrano to run the migrations on deploy?

Deferring to @aaron-collier for the approval.

README.md
```diff
@@ -109,6 +109,29 @@ The `optimistic_locking_management` DAG requires a Postgres Airflow
...
+- **DATABASE_HOSTNAME**
+
+(If you're having problems setting these with the `.env` file when running locally, add the keyword `export` before each variable in that `.env` file 
+and then run `source .env`)
+
+To generate a migration script, first make the changes in the `models.py` 
+module and then run the following steps:
+ 
+ 1. From the root directory, run `alembic revision --autogenerate -m "{short message describing change}"` (**NOTE:** not all changes to the model are detected, see this [note](https://alembic.sqlalchemy.org/en/latest/autogenerate.html#what-does-autogenerate-detect-and-what-does-it-not-detect) in the documentation)
+ 1. After the migration script is created, run `alembic upgrade head` to apply your latest changes to the database.
```
aaron-collier, line 133: Note: I had to run this as `poetry run alembic upgrade head`, I'm assuming based on running in a venv maybe, should that be captured here specifically or elsewhere in the README like "if you have error running commands, try poetry run..."

Thanks.

aaron-collier (approved):

README.md
jermnelson: I updated to use the `poetry run` syntax. In the next section on **Testing** it includes a comment to use `poetry shell` to activate the venv for running alembic directly.

jermnelson:
> This looks good to me, and I'm glad you were able to get Alembic to manage migrations without requiring the dropping of data!
> 
> I liked the idea of colocating the db configuration and migrations with the vendor load app itself. But I can live with them being separate. I'm confused, don't the migrations need to run in the deployed environments (e.g. prod) and thus need to be copied into the Docker container so that they can be run from in there?
> 
> I think one follow on ticket from this (if we don't have it already) is to get Capistrano to run the migrations on deploy?
> 
> Deferring to @aaron-collier for the approval.

I believe that since Capistrano copies the entire repository, the migrations would be run through Capistrano and not in the containers themselves during the deploy.

edsu:
Got it. I guess for Capistrano to run the migrations it will need a Python Poetry environment then? Maybe it already does?

PR #368, opened 2023-05-04
add files table to database

### LINKED ISSUE #334: Initialize file model for tracking vendor data load status and history
Using ~~https://github.com/sul-dlss/libsys-airflow/pull/329/files~~ PR #333 as an example, initialize the "files" model based on https://docs.google.com/document/d/1xjbxiFWHT3THgaYC6mP1MS6VKoQKA2GWKey1TE0Y5cE/edit#heading=h.6qi50lyx2uxz

(db entity diagram at https://github.com/sul-dlss/libsys-airflow/wiki/Vendor-Load:--Database-Model)

This model is required to capture filenames from vendors with size and date attributes to determine when a file is "new" or has been loaded.

- [ ] Add files model under the data management plugin
- [ ] Update readme (or wiki) to include how to use the model with local database vendor_interface id to determine if a file is new

ndushay:
REPLACED by #375

Closes #334

Questions:
1.  is the enum for status correct?   Note that my research implies we can add values but it is a pain to delete them.
    - Aaron says yes.
2. I couldn't figure out how to initialize the created timestamp via the server's "now" function  (for created timestamp) -- perhaps this can be a separate ticket.
    - ~~I will create a separate ticket.~~  See issue #373
3. I didn't see a way to have the server use the server's "Now" function onupdate (for updated timestamp) -- perhaps this can be a separate ticket.
    - ~~I will include this in the ticket for the previous point.~~  See issue #373
4. should ticket(s) be made to add created and updated datetime fields to vendors and vendor_interfaces tables?
    - Aaron will think through this today
5. it's possible that "default" is redundant when there is also "server_default"
    - ~~I will include this in the ticket for points 2 and 3.~~  See issue #373
6. should timestamps be in UTC or local time?  should they indicate timezone??
    - Aaron will try to answer this soon.  Mentioned in issue #373
7. should vendor_interface table have last_folio_update column?
    - Aaron says "I don't think so at this time. I'm not sure what value we would get from this as we're not necessarily tracking changes upstream in folio, if we need to. I will ticket this separately."

(@aaron-collier note questions ^^)

After applying this change as written:

```
$ docker exec -it libsys-airflow-postgres-1 /bin/bash
root@7da52e512512:/# psql -U airflow
psql (14.7 (Debian 14.7-1.pgdg110+1))
Type "help" for help.

airflow=# \c vendor_loads
You are now connected to database "vendor_loads" as user "airflow".
vendor_loads=# \d
                   List of relations
 Schema |           Name           |   Type   |  Owner  
--------+--------------------------+----------+---------
 public | alembic_version          | table    | airflow
 public | files                    | table    | airflow
 public | files_id_seq             | sequence | airflow
 public | vendor_interfaces        | table    | airflow
 public | vendor_interfaces_id_seq | sequence | airflow
 public | vendors                  | table    | airflow
 public | vendors_id_seq           | sequence | airflow
(7 rows)

vendor_loads=# \d+ files
                                                                        Table "public.files"
       Column        |            Type             | Collation | Nullable |              Default              | Storage  | Compression | Stats target | Description 
---------------------+-----------------------------+-----------+----------+-----------------------------------+----------+-------------+--------------+-------------
 id                  | integer                     |           | not null | nextval('files_id_seq'::regclass) | plain    |             |              | 
 created             | timestamp without time zone |           | not null |                                   | plain    |             |              | 
 updated             | timestamp without time zone |           | not null |                                   | plain    |             |              | 
 vendor_interface_id | integer                     |           |          |                                   | plain    |             |              | 
 vendor_filename     | character varying(250)      |           | not null |                                   | extended |             |              | 
 filesize            | integer                     |           | not null |                                   | plain    |             |              | 
 vendor_timestamp    | timestamp without time zone |           |          |                                   | plain    |             |              | 
 loaded_timestamp    | timestamp without time zone |           |          |                                   | plain    |             |              | 
 status              | filestatus                  |           | not null | 'not_fetched'::filestatus         | plain    |             |              | 
Indexes:
    "files_pkey" PRIMARY KEY, btree (id)
Foreign-key constraints:
    "files_vendor_interface_id_fkey" FOREIGN KEY (vendor_interface_id) REFERENCES vendor_interfaces(id)
Access method: heap

vendor_loads=# exit
```


aaron-collier:
@ndushay attempt to answer questions:

1.) Yes, that seems correct. 
2.) I'm not sure how to do this in the database as a default. I did it in code by importing [datatime](https://github.com/sul-dlss/libsys-airflow/blob/9a23725c33a692c00cb5df3c382511fc0b6981c2/bin/seed_vendors#L17) and then setting the value to [datatime.now()](https://github.com/sul-dlss/libsys-airflow/blob/9a23725c33a692c00cb5df3c382511fc0b6981c2/bin/seed_vendors#L78) when creating the object.
3.) hmm... no idea.
4.) Perhaps, I'll think through this today
5.) I'm not sure I follow. This probably shouldn't block this PR but we can discuss and ticket updates.
6.) I will try to answer that better soon. 
7.) I don't think so at this time. I'm not sure what value we would get from this as we're not necessarily tracking changes upstream in folio, if we need to. I will ticket this separately.

ndushay:
I put Aaron's answers inline in the description.   This is now ready for review and merge.  I will write the separate ticket I volunteered for now.

ndushay:
I'm going to rename this table "vendor_files" for better clarity.  Yay me.

PR #369, opened 2023-05-05
Adds task for batching MARC files.

### LINKED ISSUE #255: Split MARC BIB Record loads into Batchs for Ingestion
Following [documention](https://wiki.folio.org/pages/viewpage.action?pageId=96415164) on size of MARC BIB record loading using `mod-data-import` for Nolana, split large files into a [configuration-based]

Note: there is a MAX_ENTITIES variable used for migration that we should utilize there to set the size of the batches: https://github.com/sul-dlss/libsys-airflow/search?q=MAX_ENTITIES

(https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/variables.html) size (suggest 500) batches for loading.

justinlittman:
closes #255


libsys_airflow/plugins/vendor/marc.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def batch(download_path: str, filename: str, max_records: int) -> list[str]:
+    """
+    Splits a MARC file into batches.
+    """
+    records = []
+    index = 1
+    batch_filenames = []
+    with open(os.path.join(download_path, filename), "rb") as fo:
```
jermnelson, line 77: Not a blocker. My preference is to use to use `pathlib` over `os.path.join` as more modern idiomatic Python. 

This can also be accomplished with the pathlib package something like:
```download_path = pathlib.Path(download_path)
with (download_path / filename).open("rb") as fo:
...
```

jermnelson (changes requested):
General comments about using `pathlib` instead of the`os` as `pathlib` has the `os` library functionality used in this PR.

ndushay (approved):
Looks good to me, modulo Jeremy's request.  I'm new to python and airflow tho, so not very confident in my approval.

justinlittman:
Thanks for the tips @jermnelson. Fixed.

jermnelson (approved):

PR #374, opened 2023-05-05
Import multiple files per import

justinlittman:


tests/test_data_import.py
```diff
@@ -17,12 +17,20 @@ def folio_client():
         "fileDefinitions": [
             {
                 "id": "8b5cc830-0c3d-496a-8472-3b98aa40d109",
-                "name": "0720230118.mrc",
+                "name": "0720230118_1.mrc",
```
ndushay, line 20: is this the pattern for batches of one?  that the filename will be like this?    Asking for a friend.

ndushay (commented):
I think it's reasonable to have a test showing how it works with a single file that doesn't need to be split up, and how it works with a file that does need to be split up.  I'm looking at the diff so it's perhaps already doing both and I didn't notice.

I'd also like to see that the error message for reading a "chunk" has the right goodies.

tests/test_data_import.py
justinlittman: That is the pattern for the first file in the batch.

aaron-collier (approved):

PR #376, opened 2023-05-05
Setting up tests of the app views

### LINKED ISSUE #377: Add tests of vendor management view pages
Next step as part of #332 is to set up tests of the vendor index page in the vendor_loads plugin. I have some (non-passing) starting code on this draft PR: https://github.com/sul-dlss/libsys-airflow/pull/376

lwrubel:
Add a feature test for the vendor listing page in the Vendor Loads plugin. The steps involved:

- using pytest-mock-resources to create and populate an in memory sqlite database with some Vendor fixture data
- mock Airflow's `PostgresHook` database connection to return the mocked database
- create a `test_client` fixture that creates the Airflow Flask application, adds our Vendor Loads plugin to it, and returns a web client for talking to it

I put the `test_client` fixture in its own module and loaded it in `tests.conftest` so that it's available to any other tests that might need it.

I had hoped to put the database setup and mocking into `tests.mocks` but since `pytest_mock_resources.create_sqlite_fixture()` is itself a pytest fixture and you are no longer able to call these directly, I couldn't figure out a way to do it.

Closes #377


edsu:
This is currently failing with:

```
FAILED tests/apps/vendor_app/test_vendor_management_view.py::test_vendor_views - airflow.exceptions.AirflowNotFoundException: The conn_id `vendor_loads` isn't defined
```

libsys_airflow/plugins/vendor_app/main.py
```diff
@@ -3,6 +3,8 @@
 
 from libsys_airflow.plugins.vendor_app.vendors import VendorManagementView
 
+print(f'xxx: {__name__}')
```
ndushay, line 6: do we want this, or is it cruft from debugging?

ndushay (commented):
i don't feel qualified to review this but I noticed a print statement that probably shouldn't be checked into main?

edsu:
Oh good catch, thanks @ndushay

libsys_airflow/plugins/vendor_app/main.py
edsu: Thanks for catching that -- that was me debugging, and it has been removed.

jermnelson (approved):

tests/client.py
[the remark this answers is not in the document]
edsu: Yeah, good point, especially with it being pulled in automatically. How about `test_airflow_client`?

PR #380, opened 2023-05-05
Use libsys_airflow module everywhere

edsu:
This spike adapts the Docker containers to mount the `libsys_airflow` module as a volume in `/opt/airflow/` and then uses Airflow's `dags_folder` and `plugins_folder` in the `airflow.cfg` to point to the relevant locations. It also updates the remaining places that were doing an `import plugins` to `import libsys_airflow.plugins` instead.

The benefit of this approach is that our test code can import from `libsys_airflow` just like any code in `libsys_airflow` that needs to import from the root module. We could put `dags` and `plugins` at the top level, but I think it's nice to have our airflow code in a named module?

### DIFF
--- a/airflow.cfg
+++ b/airflow.cfg
@@ -1,2 +1,6 @@
+[core]
+dags_folder = /opt/airflow/libsys_airflow/dags
+plugins_folder = /opt/airflow/libsys_airflow/plugins
+
 [webserver]
 warn_deployment_exposure = False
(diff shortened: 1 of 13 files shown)


mjgiarlo (commented):
Without much to back up the sentiment: I prefer this more explicit approach and find it in aligment with my understanding of Pythonic values? 🤷🏻‍♂️

edsu:
I think consistency is what we probably want to shoot for. Either using the `libsys_airflow` module everywhere, or getting rid of it and having `plugins` and `dags` in the root of the repo. I prefer the former since it keeps our stuff in a namespace where it is less likely to interfere with other things?

jermnelson (approved):

PR #387, opened 2023-05-08
Adds data fetch scheduler dag

justinlittman:


docker-compose.prod.yaml
```diff
@@ -78,6 +78,7 @@ x-airflow-common:
     AIRFLOW_VAR_FOLIO_USER: ${AIRFLOW_VAR_FOLIO_USER}
     AIRFLOW_VAR_FOLIO_PASSWORD: ${AIRFLOW_VAR_FOLIO_PASSWORD}
     AIRFLOW_VAR_OKAPI_URL: ${AIRFLOW_VAR_OKAPI_URL}
+    AIRFLOW_CONN_VENDOR_LOADS: postgresql+psycopg2://airflow:airflow@postgres/vendor_loads
```
aaron-collier, line 81: I think this should use the DATABASE_USERNAME, DATABASE_PASSWORD, and DATABASE_HOSTNAME env vars as seen on line 60.

justinlittman: Ooops. Cut and paste error. Fixed.

aaron-collier (approved):

ndushay (commented):
methinks we should standardize on formatting.  A lot of this is lint noise, and could get changed back by another developer.

justinlittman:
@ndushay My understanding is that the standard is to run https://github.com/psf/black for formatting.

PR #389, opened 2023-05-08
Merge conflicting migrations

edsu:
I think we have two db migrations committed that had the same parent
migration, which caused a problem when applyin the update. Running
`alembic merge heads` seemed to fix it.

https://stackoverflow.com/a/51456404/324921


ndushay:
this will be in my PR forthcoming in a few mintues.

PR #391, opened 2023-05-09
Update vendor files records upon download.

### LINKED ISSUE #384: Update the vendor download plugin to create a VendorFile record for each new file
Refs: https://github.com/sul-dlss/libsys-airflow/blob/main/libsys_airflow/plugins/vendor/download.py#L61

Update the download plugin to create a VendorFile for each new file fetched. If there was an error fetching set the status to `fetching_error` otherwise the status should be `fetched`.

justinlittman:
closes #384

### DIFF
--- a/libsys_airflow/plugins/vendor/download.py
+++ b/libsys_airflow/plugins/vendor/download.py
@@ -3,21 +3,34 @@ import re
 import pathlib
 from typing import Union
 import os
+from datetime import datetime
+
+from sqlalchemy.orm import Session
+from sqlalchemy import select
 
 from airflow.decorators import task
 from airflow.providers.ftp.hooks.ftp import FTPHook, FTPSHook
+from airflow.providers.postgres.hooks.postgres import PostgresHook
+
+from libsys_airflow.plugins.vendor.models import VendorFile, VendorInterface
 
 logger = logging.getLogger(__name__)
 
 
 @task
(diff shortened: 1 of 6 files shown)


libsys_airflow/plugins/vendor/models.py
```diff
@@ -46,6 +46,7 @@ class VendorInterface(Model):
     processing_dag = Column(String(50), unique=False, nullable=True)
     processing_delay_in_days = Column(Integer, unique=False, nullable=True)
     active = Column(Boolean, nullable=False, default=False)
+    vendor_files = relationship("VendorFile", back_populates="vendor_interface")
```
aaron-collier, line 49: Do the changes here require a migration?

libsys_airflow/plugins/vendor/download.py
```diff
@@ -38,6 +51,8 @@ def download(
     remote_path: str,
     download_path: str,
     filename_regex: str,
+    vendor_interface_uuid: str,
+    pg_hook: PostgresHook = None,
```
edsu, line 55: Is it weird that we are having to tell our non-test code about things that are required by our tests?

aaron-collier: Perhaps, but it being optional helps.

aaron-collier (approved):

PR #398, opened 2023-05-10
Handles processing delay when initiating default data processor DAG

justinlittman:


libsys_airflow/dags/vendor/data_fetch_scheduler.py
```diff
@@ -40,7 +41,7 @@ def generate_dag_run_conf() -> list[dict]:
         with Session(pg_hook.get_sqlalchemy_engine()) as session:
             for vendor_interface in (
                 session.query(VendorInterface)
-                .filter(VendorInterface.active is True)
+                .filter(VendorInterface.active == true())
```
justinlittman, line 44: This is a bug fix. Earlier I let flake lead me astray.

jermnelson (approved):

PR #400, opened 2023-05-10
Fix creation of data processor dags. Limit data import task.

justinlittman:


libsys_airflow/dags/vendor/data_fetcher.py
```diff
@@ -75,33 +77,33 @@ def setup():
         return params
 
     @task
-    def setup_dag_run_conf(
-        downloaded_files: list,
+    def generate_dag_run_kwargs(
```
justinlittman, line 80: These changes adjust the way that processing DAGs are kicked off so that the DAGs end up unique (as required by Airflow).

libsys_airflow/plugins/folio/data_import.py
```diff
@@ -31,6 +31,7 @@ def record_loading_error(context):
     on_execute_callback=record_loading,
     on_success_callback=record_loaded,
     on_failure_callback=record_loading_error,
+    max_active_tis_per_dag=Variable.get("max_active_data_import_tis", default_var=1),
```
justinlittman, line 34: This prevents slamming Folio.

ndushay (approved):

Issue #394, opened 2023-05-09
[mini epic] Improve CI builds:  use containers, add format checks, etc.


ndushay:
Per thread in slack (https://stanfordlib.slack.com/archives/C04EZLBMLQN/p1683665868232099?thread_ts=1683663418.751839&cid=C04EZLBMLQN), we want to improve CI builds:

- [ ] have our GH action use our docker containers for our CI build, as our deployments will  ( https://docs.github.com/en/actions/using-containerized-services/about-service-containers may help)
    - per standup discussion 5/16/23, we've decided this is not worth it because:
        - we believe the vendor_load database to be stable now ... no new migrations expected
        - our tests are very "unit-y" and use sqlite and mocking, so spinning up the whole megillah in docker to run the tests seems unnecessary.
    - [ ] run alembic migrations in docker containers as part of the build, even if the tests don't need them -- this is a way to ensure we will find out if the alembic/migrations history ends up with multiple head pointers, for example.
        - this should be a separate workflow (?)
        - issue to make this part of deployment:  #372
- [x] have our GH actions also do both of these (as a separate workflow):
    - [x] format checking (using `black`) - see PRs #412, #413, #414
    - [x] linting (using `flake8` or possibly replacing with `ruff`) - see PRs #423 and #424
- [x] do NOT ignore rule F401 about unused import statements. -- see PR #425
- [ ] have our formating standardize on organization of import statements in our files.  (plugin for flake8)
    - I tried `isort`, `flake8-isort` and `flake8-import-order` and they all seemed non-trivial to configure.  For example, for `flake8-isort`, I removed a blank line it complained about, and then got a flake8 error.   I even tried setting the profile to black, and it didn't help.   Thus, punting.

These could be split into separate tickets, or they could be separate PRs on this ticket.

![image.png](https://images.zenhubusercontent.com/5d89272aae87b600015057e7/5e2bb5f1-cdc1-42e9-b969-c34e0b8cc31f)

jmartin-sul:
here's an example of using `black` in CircleCI to find lint that `black` would correct: https://github.com/sul-dlss/dlme-airflow/blob/main/.circleci/config.yml#L39-L42 (the default mode is to just do the corrections, this mode returns a non-zero exit code if `black` finds things it'd change: `poetry run black --diff --check .`)

jmartin-sul:
[...]

PR #414, opened 2023-05-12
Update black github action and provide an easy way to fix black formatter violations

ndushay:
part of #394

I believe this makes us ready for prime time with `black` formatter.

It runs in check mode by default, reporting on violations.   There are instructions in the readme for how to let black automatically fix the violations.   This is akin to rubocop, in that we do not fix the violations automatically on the build;  the build notifies, and the developer fixes them.

The way I found to do this was to create a config file with the desired options to FIX the problem, and to document the command to run black using this config file as part of the README.   

NOTE:  you will see from my inline comments that this PR will not be green, but if it is merged to main, it will be green afterwards -- flake8 and black are disagreeing on a style rule and this PR addresses it.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,12 @@
+# github action workflow to run black formatter check and linting checks for our python code
+
 name: Lint
 
-on: workflow_dispatch
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
 
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -5,9 +5,9 @@ name: Folio Airflow CI
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
+    branches: [main]
 
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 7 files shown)


.github/workflows/lint.yml
```diff
@@ -1,6 +1,12 @@
+# github action workflow to run black formatter check and linting checks for our python code
+
 name: Lint
 
-on: workflow_dispatch
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
```
ndushay, line 9: this makes the black formatting happen as part of PRs and PR merges to main.

libsys_airflow/plugins/folio/instances.py
```diff
@@ -159,7 +159,7 @@ def post_folio_instance_records(**kwargs):
         instance_records = json.load(fo)
 
     for i in range(0, len(instance_records), batch_size):
-        instance_batch = instance_records[i: i + batch_size]
+        instance_batch = instance_records[i : i + batch_size]
```
ndushay, line 162: a black formatting error that was merged to main while I was working on this PR.

pyproject.toml
```diff
@@ -45,3 +45,6 @@ target-version = ['py310']
 include = '\.pyi?$'
 exclude = 'vendor_loads_migration'
 skip-string-normalization = true
+check = true
+diff = true
+color = true
```
ndushay, line 50: we want to _check_ by default, and fix on demand.

.github/workflows/python-app.yml
```diff
@@ -5,32 +5,32 @@ name: Folio Airflow CI
...
+        with:
+          python-version: '3.10'
+      - name: Install Airflow + dependencies
+        run: |
+          pip install -r requirements.txt
+          poetry install
+      - name: Lint with flake8
+        run: |
+          cd libsys_airflow
+          poetry run flake8 --ignore=E203,E225,E501,F401,F811,W503 dags/ plugins/
```
ndushay, line 30: See https://github.com/psf/black/issues/315 for why we need to tell flake to ignore E203.

jermnelson (approved):

PR #416, opened 2023-05-15
Add Vendor view

### LINKED ISSUE #355: Vendor Load UI: Add vendor show page
The initial iteration of this page should be very similar to the vendor list page, but will be listing the individual interfaces for the vendor.

Display in a simple table:

- Interface display name (will link to the vendor interface configuration/show page)
- data import profile name
- file regeex
- remote path
- processing delay
- processing dag

Note: Everything except the interface display name _can_ be blank - as they will have no values until they are configured. One approach to this UI could be to show the combination of vendor_interfaces from the database as well as any vendor_interfaces returned from FOLIO but not in the database (i.e. vendor_interfaces to be configured) or there could be a task to populate the database of vendors and interfaces and just reflect the DB.

edsu:
Add a vendor view that lists the interfaces for a given vendor. The URL path is:

    /vendors/{vendor-id}

Initially I had this as:

    /vendors/vendor/{vendor-id}

But it looked kind of weird? We can change it later if it causes problems with subsequent views?

I moved `tests.client` to `tests.airflow_client` to avoid confusion with other clients. I also modified the test_airflow_client response so that it can provide access to a BeautifulSoup object for the response to allow the tests to operate on the parsed HTML rather than doing straight up string matching.

Closes #355

### DIFF
--- a/libsys_airflow/plugins/vendor_app/templates/vendors/index.html
+++ b/libsys_airflow/plugins/vendor_app/templates/vendors/index.html
@@ -12,7 +12,7 @@
         {% for vendor in vendors %}
         <tr>
             <td>
-                {{ vendor.display_name }}<br>
+              <a href="{{ url_for('VendorManagementView.vendor', vendor_id=vendor.id) }}">{{ vendor.display_name }}</a><br>
                 ID: {{ vendor.folio_organization_uuid }}
             </td>
             <td>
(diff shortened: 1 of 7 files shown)


edsu:
Hmm, it doesn't look like our CI tests are running anymore?

aaron-collier (approved):

jermnelson (approved):

PR #419, opened 2023-05-15
Use dotenv

edsu:
Use python-dotenv to simplify the running of `seed_vendors`, assuming that the `.env` file is already set up (development), or the `AIRFLOW_VAR_OKAPI_URL`, `AIRFLOW_VAR_FOLIO_USER` and `AIRFLOW_VAR_FOLIO_PASSWORD` environment variables have been set up by Puppet (qa, stage, prod).

### DIFF
--- a/bin/seed_vendors
+++ b/bin/seed_vendors
@@ -3,12 +3,11 @@
 # A utility for getting a vendor records from FOLIO into the database
 
 import os
-import sys
 import logging
 import argparse
 import requests
-import json
 
+from dotenv import load_dotenv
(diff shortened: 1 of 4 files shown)


bin/seed_vendors
```diff
@@ -3,19 +3,20 @@
 # A utility for getting a vendor records from FOLIO into the database
 
 import os
-import sys
 import logging
 import argparse
 import requests
-import json
```
edsu, line 10: These were removed since they weren't being used.

justinlittman (approved):

PR #420, opened 2023-05-15
Update default_view for vendors

lwrubel:
Bugfix related to shift in view method.


edsu:
Thanks for fixing this!

PR #421, opened 2023-05-15
Adds support for SFTP downloads.

### LINKED ISSUE #408: Support SFTP download
Currently, only FTP download is supported; Amalivre and MarcIt require SFTP.

Note:
* Creating SFTP connections is supported: https://github.com/sul-dlss/libsys-airflow/pull/402
* Marcit is IP restricted, so testing is problematic.

justinlittman:
closes #408

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ pytest-mock = "^3.7.0"
 # the airflow version here should be kept in sync with the docker image
 apache-airflow = "2.6.0"
 apache-airflow-providers-postgres = "5.4.0"
+apache-airflow-providers-sftp = "^4.2.4"
 requests-mock = "^1.10.0"
 pytest-mock-resources = "^2.6.12"
 beautifulsoup4 = "^4.12.2"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -22,6 +22,7 @@ apache-airflow-client = "2.6.0"
 python-dotenv = "^1.0.0"
 alembic = "^1.10.4"
 psycopg2-binary = "^2.9.6"
+apache-airflow-providers-sftp = "^4.2.4"
```
edsu, line 25: I believe that the Docker Image pulls this in already as an [extra package](https://airflow.apache.org/docs/apache-airflow/stable/extra-packages-ref.html)? One way of confirming this would be too try importing it on one of our deployed Docker environments. I think we may want to be careful about installing a version on top of the one that we are getting with the image?

One thing you might want to consider is adding it to the `test` section so that we only install it in our non Docker development environments? Admittedly this all feels kind of fragile, and perhaps we need to bite the bullet and just start using the Docker image for development?

edsu (approved):
Good stuff! I left some comments for consideration.

pyproject.toml
justinlittman: Moving to test section.

PR #429, opened 2023-05-15
Utility for getting job profiles

### LINKED ISSUE #406: Add a utility to libsys/plugins/vendor to query data import profiles from FOLIO
When configuring vendor_interfaces for data importing, we need to have a select list of all data import profiles to allow the selection of which one to link to the vendor_interface.

Add a utility that uses the folio client to query all data import profiles and returns a list/dict/??? so that we have the name for display in the interface and the data import profile UUID to apply to the vendor_interface.

lwrubel:
Resolves #406 to get a list of data import job profiles.


libsys_airflow/plugins/vendor/job_profiles.py
```diff
@@ -0,0 +1,23 @@
...
+            Variable.get("FOLIO_USER"),
+            Variable.get("FOLIO_PASSWORD"),
+        )
+
+    job_profiles_resp = folio_client.get("/data-import-profiles/jobProfiles")
+
+    return [
+        {"id": profile["id"], "name": profile["name"]}
+        for profile in job_profiles_resp["jobProfiles"]
+    ]
```
edsu, line 23: I like the use of a list comprehension here. Is there a reason not to return all the data instead of cherry picking these two values?

edsu (approved):
I left a question, not a request for changes though.

libsys_airflow/plugins/vendor/job_profiles.py
lwrubel: That could totally work, and is probably more efficient? Since all that is needed for the UI dropdown is the id and name, and it's a level nested in the JSON response, I thought I'd make the structure simpler for eventual use in the view. I also realized, although it's not mentioned in the issue, that these will probably need to be sorted in some way. So I just added the logic to sort by job profile name for that to this utility.

lwrubel: @edsu since I made a change since you approved, would you give it a quick glance over again before I merge?

libsys_airflow/plugins/vendor/job_profiles.py
```diff
@@ -16,8 +17,9 @@ def job_profiles(folio_client=None) -> list:
...
 
     job_profiles_resp = folio_client.get("/data-import-profiles/jobProfiles")
-
-    return [
+    job_profiles = [
         {"id": profile["id"], "name": profile["name"]}
         for profile in job_profiles_resp["jobProfiles"]
     ]
+
+    return sorted(job_profiles, key=itemgetter("name"))
```
edsu, line 25: Nice, today I learned about `operator.itemgetter` :-)

edsu (approved):

PR #435, opened 2023-05-16
Adding folio_job_execution_uuid to FileDataLoad

### LINKED ISSUE #426: Extend the FileDataLoad model to include folio job_execution_id
When a data_import (https://github.com/sul-dlss/libsys-airflow/blob/main/libsys_airflow/plugins/folio/data_import.py#L49) is initiated we need to capture the job_execution_id as this will be used to link back directly to folio for task status and reporting.

Add `folio_job_execution_id` to the FileDataLoad model.

lwrubel:
Resolves #426 to add `folio_job_execution_uuid` as a field to the `FileDataLoad` model.


libsys_airflow/plugins/vendor/models.py
```diff
@@ -110,6 +110,7 @@ class FileDataLoad(Model):
         server_default=DataLoadStatus.not_loaded.value,
     )
     additional_info = Column(Text, nullable=True)
+    folio_job_execution_uuid = Column(String(36), unique=False, nullable=True)
```
lwrubel, line 113: I called the field `folio_job_execution_uuid` instead of `folio_job_execution_id` (as the issue requested) because we have a pattern of naming folio fields with `uuid` in the models.

edsu (approved):

PR #442, opened 2023-05-16
Plugin method to move marc fields

### LINKED ISSUE #431: Add a method to plugins.vendor.marc to move marc fields in a given file
We must move some marc fields from vendors, for instance change the 520 to 920 and 528 to 928. 

Given the path to a marc file and `from` and `to` values change the marc fields in ever record in the file.

lwrubel:
Resolves #431 to provide method to move marc fields in a file.

### DIFF
--- a/libsys_airflow/plugins/vendor/marc.py
+++ b/libsys_airflow/plugins/vendor/marc.py
@@ -118,3 +118,42 @@ def _write_records(records, download_path, filename):
         for record in records:
             record.force_utf8 = True
             marc_writer.write(record)
+
+
+def move_fields(marc_path: pathlib.Path, change_fields: list[dict]):
+    """
+    Moves MARC fields within each record in a file.
+    """
+    if not all(change.keys() == {"from", "to"} for change in change_fields):
+        logger.error(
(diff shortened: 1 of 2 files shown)


libsys_airflow/plugins/vendor/marc.py
```diff
@@ -118,3 +118,42 @@ def _write_records(records, download_path, filename):
         for record in records:
             record.force_utf8 = True
             marc_writer.write(record)
+
+
+def move_fields(marc_path: pathlib.Path, change_fields: list[dict]):
+    """
+    Moves MARC fields within each record in a file.
+    """
+    if not all(change.keys() == {"from", "to"} for change in change_fields):
```
lwrubel, line 127: Not sure where is best to be validating the changes dictionary, or how these will be created, so putting it here for now. Open to other ideas.

aaron-collier: Validation did cross my mind but didn't want to confuse this ticket. I think this is good for now, but may need to be addressed if we add more.

aaron-collier (approved):

PR #445, opened 2023-05-17
Improved data type detection.

justinlittman:
Turns out accurate date type detection is needed for import.


edsu (changes requested):
I wonder if we should maybe consider switching to puremagic instead so we don't need to fuss with the system dependencies?

edsu:
Nevermind, puremagic doesn't seem to recognize MARC data :-( -- just does files with magic numbers I guess.

edsu (approved):

justinlittman:
Thanks for the suggestion and investigation @edsu!

Issue #411, opened 2023-05-11
Airflow Task: Change marc field codes


aaron-collier:
In several cases, we need to change a set of marc fields to a different code. For example, with Amalivre, we change the 505 to 905 and the 520 to 920.

Get the field-to-field list from the interface configuration.

More to come.

PR #446, opened 2023-05-17
VendorInterface model and db table: add processing_options field

### LINKED ISSUE #430: Add processing_options to VendorInterface model
In some cases, we need the option to configure how tasks are run during processing. This will include optionally moving one marc code to another (i.e. 520 moved to 920).  In order to capture this data in the model, we'll use a hash, something like:

```
{ change_marc: [
  { from: 520 to: 920 },
  { from: 528 to: 928 }
 ]}
```

Note: that will be specified in a follow up ticket, this is just an example that will allow us to avoid continuously adding single fields for configuration options.

ndushay:
Closes #430 

The new field will be used by #411 to add additional processing information to vendor interfaces (e.g. copy marc 505 to 905)


justinlittman:
Do we want this to be JSONB instead of text?

ndushay:
> Do we want this to be JSONB instead of text?

Good idea.

ndushay:
@aaron-collier so we're wondering 

- should this field be JSON?  
- whether we care about JSONB ... will we want to query this field?
- how sqlalchemy will handle database field types - it "seems" to be ok with JSON per https://docs.sqlalchemy.org/en/20/core/types.html ... but 
- how we should test how the field type is working ... without having specific requirements of what JSON needs to go in the field, and more info on how it will be utilized (e.g. before other marc tranforms?   after?  instead of?   in which DAG?  etc.)
- whether there are implications that our test database should be postgres, in case there are differences in how JSON is handled between postgres and sqllite

etc.

justinlittman:
I added on to this PR to add the move MARC fields task to the default data processor dag.

ndushay:
I'm not clear if moving marc fields will always occur before other marc processing, or always occur after, or sometimes before and sometimes after.  If it's consistent then presumably this PR puts it in the right spot in the DAG.

justinlittman:
I'm unaware of any cases in which the order matters; if we encounter such a case, we can reconsider.

libsys_airflow/plugins/vendor/marc.py
```diff
@@ -121,17 +122,36 @@ def _write_records(records, download_path, filename):
...
+        logger.info(f"Skipping moving fields for {marc_path}")
+        return
+    move_fields(pathlib.Path(marc_path), change_fields)
+
+
 def move_fields(marc_path: pathlib.Path, change_fields: list[dict]):
     """
     Moves MARC fields within each record in a file.
+
+    change_fields example: [{ from: 520 to: 920 }, { from: 528 to: 928 }]
```
lwrubel, line 146: Glad you added this example. Pymarc is expecting a tag to be a string, not an integer, so the example and code should reflect this. 

Example would be something like:

`[{ "from": "520", "to": "920" }, { "from": "528", "to": "928" }]`

@aaron-collier and I talked a little about validation of change fields and agree it may be needed beyond what is in line 148, but will handle when needed and/or we understand better the kinds of changes needed to be implemented.

justinlittman: Fixed example.

lwrubel (approved):

PR #456, opened 2023-05-18
Multiple minor DAG-related tweaks

justinlittman:


libsys_airflow/plugins/airflow/connections.py
```diff
@@ -48,7 +48,7 @@ def find_or_create_conn(
     key_file: Optional[str] = None,
     session: Session = NEW_SESSION,
 ) -> str:
-    conn_id = f"{conn_type}-{host}"
+    conn_id = f"{conn_type}-{host}-{login}"
```
justinlittman, line 51: Turns out we had multiple connections to the same FTP host with different credentials. This makes adequately unique.

jmartin-sul (approved):

PR #471, opened 2023-05-20
query DAGs by tag

### LINKED ISSUE #410: Vendor Load UI: Add utility to libsys/plugins/vendor to query DAGs by tag: "vendor load"
Add a utility that queries the airflow API for DAGs where the tag is "vendor load" so that the list of available processing dags can be included in the vendor_interface configuration UI.

jmartin-sul:
closes #410

this is a slightly different approach than what the original ticket called for, but that decision came out of post-standup discussions and pairing, so hopefully that's still ok.

i tried to note gotchas to this approach in the code comments.

### DIFF
--- a/libsys_airflow/plugins/airflow/airflow_queries.py
+++ b/libsys_airflow/plugins/airflow/airflow_queries.py
@@ -1,0 +1,24 @@
+import logging
+
+from airflow.models import DagBag, DAG
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_dags(filter_tags: set[str]) -> set[DAG]:
+    """
+    Return the set of DAGs that match at least one of the filter criteria.
+    * Approach lightly adapted from https://stackoverflow.com/questions/49983034/get-dag-from-airflow (which lists all the tasks in all the DAGs).
+    * An alternate approach would be to use the REST API via apache-airflow-client, since it offers a way to list DAGs
+      filtered by tag, see https://airflow.apache.org/docs/apache-airflow/2.6.0/stable-rest-api-ref.html#operation/get_dags (that
+      would require settling on a REST API authN approach).
+    * Warning: When DagBag().dags is called, warnings are logged about DAGs and tasks already being loaded, but manual
+      testing indicates that no trouble is caused by the call (i.e. all the DAGs still show up in the web UI).  This should
+      be fine if we only read from DagBag().dags.  We should likely avoid ever directly setting DagBag().dags ourselves, or we
+      should at least test any such approach thoroughly and see if Airflow docs or source indicate against such usage.
+    * dag.tags is list[str], hence the cast to set.
+    """
+    return set(
+        [dag for dag in DagBag().dags.values() if len(set(dag.tags) & filter_tags) > 0]
+    )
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -17,8 +17,6 @@ pymarc = "^4.1.3"
 folio-migration-tools = "1.7.8"
 requests = "^2.28.2"
 pyyaml = "^6.0"
-# this should ideally be kept in sync with the airflow version but client releases run behind server releases
-apache-airflow-client = "2.6.0"
```
jmartin-sul, line 21: @jermnelson and i discussed removing this dependency when we paired on this work yesterday, since it wasn't clear that anything else would need the REST client, and since nothing is using it at present (the original approach discussed for this ticket was to use the REST API).  but jeremy suggested running by @aaron-collier first for approval.  i did this as a second commit, so very easy to lop off if we want to keep this dep.

tests/test_airflow_queries.py
```diff
@@ -0,0 +1,94 @@
...
+    assert get_dag_ids(get_dags(filter_tags=set(['dataset-scheduled']))) == set(
+        [
+            'dataset_produces_2',
+            'dataset_consumes_unknown_never_scheduled',
+            'dataset_consumes_1_and_2',
+            'dataset_consumes_1_never_scheduled',
+            'dataset_consumes_1',
+            'dataset_produces_1',
+        ]
+    )
```
jmartin-sul, line 29: using sets made it easier to compare lists without regard to order, and i don't think we'd expect duplicates?

tests/test_airflow_queries.py
```diff
@@ -0,0 +1,94 @@
+from airflow.models import DAG
+from libsys_airflow.plugins.airflow.airflow_queries import get_dags
+
+
+def get_dag_ids(dags: set[DAG] | list[DAG]) -> set[str]:
+    return set([dag.dag_id for dag in dags])
```
jmartin-sul, line 6: possibly overkill to have this helper, but i felt like it made the other test code more readable.

libsys_airflow/plugins/airflow/airflow_queries.py
```diff
@@ -0,0 +1,24 @@
...
+      filtered by tag, see https://airflow.apache.org/docs/apache-airflow/2.6.0/stable-rest-api-ref.html#operation/get_dags (that
+      would require settling on a REST API authN approach).
+    * Warning: When DagBag().dags is called, warnings are logged about DAGs and tasks already being loaded, but manual
+      testing indicates that no trouble is caused by the call (i.e. all the DAGs still show up in the web UI).  This should
+      be fine if we only read from DagBag().dags.  We should likely avoid ever directly setting DagBag().dags ourselves, or we
+      should at least test any such approach thoroughly and see if Airflow docs or source indicate against such usage.
+    * dag.tags is list[str], hence the cast to set.
+    """
+    return set(
+        [dag for dag in DagBag().dags.values() if len(set(dag.tags) & filter_tags) > 0]
```
jmartin-sul, line 23: another approach would be to use `filter` and provide the `if` logic as a lambda, but this list comprehension approach felt more idiomatic/pythonic?  happy to do it the other way if that's preferable.

jmartin-sul (commented):
questions and explanatory comments inline

jermnelson (approved):

aaron-collier (approved):

PR #483, opened 2023-05-22
Use simpledatatables for Vendor Pagination

### LINKED ISSUE #378: Vendor list is paged and searchable by name and code
Follow-on to #332 for making the vendor list ready for larger lists of vendors. Vendor list (/vendors) should be paged and searchable by name and code.

I thought https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/pagination/ might be useful, but it requires using [Flask-SQLAlchemy](https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/), which we are not currently using. We're using [SQLAlchemy ](https://www.sqlalchemy.org/) directly. It may be worth investigating implementing Flask-SQLAlchemy to make some of the UI tasks simpler.

jermnelson:
Fixes #378

This draft PR is mostly for discussion on adding support for pagination to the vendor tables in Vendor's view using client-side Javascript  [simple-datatables](https://github.com/fiduswriter/simple-datatables) library. To test the viability of this approach, all 3,258 Organizations from FOLIO test were loaded into the local Postgres database. 

When accessing the `/vendors` page, all 3K+ organizations are rendered in DOM and then the simple-datatables's DataTable javascript is applied to the table providing functional pagination and searching across the three columns. Loading and rendering all of these organizations into the DOM took about 550 ms with Javascript scripting taking 1348 ms for a total of 1.89 seconds to fully render to the user, which is under the recommend maximum rendering time of 2 seconds. 
  
Screenshot 2023-05-22 at 3 52 21 PM

## PROS
- Very easy implementation
- Javascript library provides what we need with little customization 
- Likelihood of significant increase of organizations low 
- Degrades gracefully if Javascript isn't loaded

## CONS
- Close to threshold of loading usability of 2 seconds (may be longer when run over the network)
- Entire DOM is rendered server-side (i.e. big "chunk" of HTML)
- Not scalable if number of organizations significantly increase

### TODO
- [x] Use Bootstrap styling for **entries per page** drop-down select and **Search..** box
- [x] Pin simple-datatables to specific version with CDN

### DIFF
--- a/libsys_airflow/plugins/vendor_app/templates/vendors/index.html
+++ b/libsys_airflow/plugins/vendor_app/templates/vendors/index.html
@@ -2,12 +2,12 @@
 
 {% block head_css %}
     {{ super() }}
-    <link href="https://cdn.jsdelivr.net/npm/simple-datatables@latest/dist/style.css" rel="stylesheet" type="text/css">
+    <link href="https://cdn.jsdelivr.net/npm/simple-datatables@7/dist/style.css" rel="stylesheet" type="text/css">
 {% endblock %}
 
 {% block head_js %}
     {{ super() }}
-    <script src="https://cdn.jsdelivr.net/npm/simple-datatables@latest" type="text/javascript"></script>
+    <script src="https://cdn.jsdelivr.net/npm/simple-datatables@7" type="text/javascript"></script>
 {% endblock %}
 
 {% block content %}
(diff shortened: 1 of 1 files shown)


mjgiarlo:
This is a compelling analysis. Given the alternatives, I'd be inclined to go with this implementation over one that involves extensive re-engineering (_e.g._, layering Flask-SQLAlchemy over the current models to get access to its pagination functionality), until we are reasonably confident that 1. the number of vendors is going to grow significantly in the near-term, or 2. the performance is preventing actual users from getting their work done.

lwrubel (approved):

PR #488, opened 2023-05-24
Add ability to disable and enable vendor interfaces in UI

### LINKED ISSUE #479: Add ability to enable/disable vendor_interface from the UI
On the vendor page, add a toggle to the list of vendor interfaces that is enabled if the interface is active, disabled if inactive and is used to set the value of the interface.

This will enable the user the ability to turn on and off vendors interface as they are being worked on/debugged/etc.

mjgiarlo:
Fixes #479

This commit changes the vendor load application so that users can disable and enable vendor interfaces, both from the vendor show view and the vendor interface edit view.

enable_disable_vendor_interfaces

### DIFF
--- a/libsys_airflow/plugins/vendor/models.py
+++ b/libsys_airflow/plugins/vendor/models.py
@@ -27,7 +27,9 @@ class Vendor(Model):
     acquisitions_unit_from_folio = Column(String(36), unique=False, nullable=False)
     has_active_vendor_interfaces = Column(Boolean, nullable=False, default=False)
     last_folio_update = Column(DateTime, nullable=False)
-    vendor_interfaces = relationship("VendorInterface", back_populates="vendor")
+    vendor_interfaces = relationship(
+        "VendorInterface", back_populates="vendor", order_by='VendorInterface.id'
+    )
 
     def __repr__(self) -> str:
         return f"{self.display_name} - {self.folio_organization_uuid}"
--- a/libsys_airflow/plugins/vendor_app/vendors.py
+++ b/libsys_airflow/plugins/vendor_app/vendors.py
@@ -63,26 +63,29 @@ class VendorManagementView(BaseView):
         VendorInterface object.
         """
 
-        if form['folio-data-import-profile-uuid']:
+        if 'folio-data-import-profile-uuid' in form.keys():
             interface.folio_data_import_profile_uuid = form[
                 'folio-data-import-profile-uuid'
             ]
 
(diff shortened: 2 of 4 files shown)


libsys_airflow/plugins/vendor/models.py
```diff
@@ -27,7 +27,7 @@ class Vendor(Model):
     acquisitions_unit_from_folio = Column(String(36), unique=False, nullable=False)
     has_active_vendor_interfaces = Column(Boolean, nullable=False, default=False)
     last_folio_update = Column(DateTime, nullable=False)
-    vendor_interfaces = relationship("VendorInterface", back_populates="vendor")
+    vendor_interfaces = relationship("VendorInterface", back_populates="vendor", order_by='VendorInterface.id')
```
mjgiarlo, line 30: This is a quality of life change: return a vendor's interfaces in a predictable order. (If we should be ordering in another way, or returning them in a random order, I'm happy to change this.)

libsys_airflow/plugins/vendor_app/vendors.py
```diff
@@ -63,26 +63,29 @@ def _update_vendor_interface_form(self, interface, form):
...
 
-        if form['remote-path']:
+        if 'remote-path' in form.keys():
             interface.remote_path = form['remote-path']
 
-        if form['file-pattern']:
+        if 'file-pattern' in form.keys():
             interface.file_pattern = form['file-pattern']
 
-        if form['package-name']:
```
mjgiarlo, line 85: Without this change, the endpoint returns an HTTP 400 because the `MultiDict` instance (`form`) will raise `KeyError` if the field specified in the form (in the `form['field_name_here']` syntax) does not exist.

libsys_airflow/plugins/vendor_app/vendors.py
```diff
@@ -63,26 +63,29 @@ def _update_vendor_interface_form(self, interface, form):
         VendorInterface object.
         """
 
-        if form['folio-data-import-profile-uuid']:
+        if 'folio-data-import-profile-uuid' in form.keys():
```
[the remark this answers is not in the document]
mjgiarlo: @edsu I initially tried `has_key()` but 🚫🎲:

Screenshot from 2023-05-24 08-24-20

As best I can tell, it was removed during the Python3-ification process? These are the methods I saw available: https://werkzeug.palletsprojects.com/en/1.0.x/datastructures/#werkzeug.datastructures.MultiDict

PR #489, opened 2023-05-24
Support upload only vendor interfaces.

### LINKED ISSUE #481: Add ability to add an "upload only" interface to a vendor
For vendors that do not have configured interfaces for automated data retrieval in FOLIO, we will need to establish a "upload only" interface in order to both maintain the selected processing configuration but also to establish where files are stored that match our .../vendor_id/interface_id pattern.

- Generate a UUID for the folio_interface_uuid (or add a second ID field if using folio_interface_uuid would complicate things)
- Set display name to "Upload Only"
- Do not allow "Upload Only" interfaces to become "active" to avoid errors in automation DAGs

justinlittman:
closes #481

image


libsys_airflow/dags/vendor/data_fetch_scheduler.py
```diff
@@ -42,6 +42,7 @@ def generate_dag_run_conf() -> list[dict]:
             for vendor_interface in (
                 session.query(VendorInterface)
                 .filter(VendorInterface.active == true())
+                .filter(VendorInterface.folio_interface_uuid != None)
```
justinlittman, line 45: So that upload only vendor interfaces aren't scheduled.

libsys_airflow/dags/vendor/data_fetcher.py
```diff
@@ -49,7 +49,7 @@
         "dataload_profile_uuid": Param(
             "", type="string"
         ),  # f4144dbd-def7-4b77-842a-954c62faf319
-        "remote_path": Param("", type="string"),  # 'oclc'
+        "remote_path": Param(None, type=["null", "string"]),  # 'oclc'
```
justinlittman, line 52: Unrelated problem that I encountered when testing.

libsys_airflow/plugins/vendor/download.py
```diff
@@ -91,7 +91,7 @@ def ftp_download_task(
 
     return download(
         hook,
-        remote_path,
+        remote_path or "",
```
justinlittman, line 94: Same.

edsu (approved):

PR #482, opened 2023-05-22
Vendor Interface edit form

edsu:
Add a view at `/vendors/inteface/{interface-id}/edit` for editing vendor interfaces.

Resolves #404

PR #492, opened 2023-05-24
Don't display None in form

edsu:
A small change to prevent the string "None" from displaying when there is no value for a VendorInterface property. Somehow this change didn't make it into #482

### DIFF
--- a/libsys_airflow/plugins/vendor_app/templates/vendors/interface-edit.html
+++ b/libsys_airflow/plugins/vendor_app/templates/vendors/interface-edit.html
@@ -52,13 +52,13 @@
   <div class="form-group">
     <label for="remote-path" class="col-sm-2 col-form-label">Remote Path</label>
     <div class="col-sm-8">
-      <input type="text" class="form-control" id="remote-path" name="remote-path" placeholder="" value="{{ interface.remote_path }}">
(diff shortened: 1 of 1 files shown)


libsys_airflow/plugins/vendor_app/templates/vendors/interface-edit.html
```diff
@@ -52,13 +52,13 @@ <h1>Edit Vendor Interface</h1>
   <div class="form-group">
     <label for="remote-path" class="col-sm-2 col-form-label">Remote Path</label>
     <div class="col-sm-8">
-      <input type="text" class="form-control" id="remote-path" name="remote-path" placeholder="" value="{{ interface.remote_path }}">
+      <input type="text" class="form-control" id="remote-path" name="remote-path" placeholder="" value="{{ interface.remote_path if interface.remote_path != None }}">
```
justinlittman, line 55: Will this work with just `if interface.remote_path`?

edsu: It does, thanks!

ndushay:
a screen shot with the change (and without) would be welcome on this PR.

justinlittman (approved):

PR #493, opened 2023-05-24
Run alembic merge heads

aaron-collier:
When running the most recent migrations I got the "multiple revisions for head" error. This seems to be the documented process to run:

`alembic merge heads`

After that, the migrations where successful. I'm not sure it makes sense to commit this, probably does, but if not be aware that locally merging heads may be required.


mjgiarlo (approved):

ndushay:
This is exactly right.  I hope we don't have too many more of these.

PR #499, opened 2023-05-25
Set VendorFile.expected_execution in UI

### LINKED ISSUE #480: Add ability to set the processing date on pending data loads
Some data loads will show as pending for several days (i.e. they are not started immediately), often this delay is arbitrary as a way to capture full bibliographic data before order data is loaded for example. If a user determines that more time is necessary to load bib data before a pending order load is processed, they need the ability to extend the processing delay by setting the date to run the processing (this could also be used to trigger processing a head of the planned delay).

edsu:
Add a view at `/vendors/file/{vendor-file-id}` that will present information about a VendorFile and allow the Expected Load Time to be updated. The input uses [&lt;input type="datetime-local"&gt;](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local) in browsers instead of jQuery plugin (which was an option that I avoided).

The list of pending files in the Interface and Dashboard views now includes the VendorFile ID which lets you navigate to the new view.

A new `folio_load_scheduler` DAG was introduced which uses the `VendorFile.ready_for_folio_load()` class method to to look for VendorFiles that are ready for loading (expected_load_time is in the past, and the status is `fetched`). It is scheduled to run every hour and starts up `default_data_processor` DAGs for any files that it finds. The dynamic execution of `default_data_processor` DAGs was removed from the `data_fetcher` DAG. The `bin/files` program lets you see what files are ready for loading on the command line.

https://github.com/sul-dlss/libsys-airflow/assets/33829/c8b30adf-bbbd-420c-94ee-6e2440d69d14

Resolves #480


libsys_airflow/plugins/vendor_app/vendors.py
```diff
@@ -165,6 +165,21 @@ def _save_file(self, vendor_uuid, interface_uuid, file_upload, execution_date):
         )
         return filepath
 
+    @expose("/file/<int:file_id>", methods=["GET", "POST"])
+    def file(self, file_id):
+        session = Session()
+        file = session.query(VendorFile).get(file_id)
+        if request.method == 'POST' and 'expected-execution' in request.form:
+            file.expected_execution = date.fromisoformat(
+                request.form['expected-execution']
```
justinlittman, line 174: This doesn't actually change the execution date. That requires changing the dag instance.

justinlittman (changes requested):

libsys_airflow/plugins/vendor_app/vendors.py
edsu: I didn't realize the information was duplicated. Is there an example of how to do that from the vendor_app yet?

edsu:
After a conversation with @justinlittman I now understand that simply updating the vendor_loads database in #499 wasn't enough to change when the DAG will actually I run, and I need to make some changes on the Airflow side of things for this rescheduling to work:

1. Move the update of `VendorFile.dag_run_id` from `libsys_airflow.dags.vendor.default_data_processor` (which currently executes as part of the data loading) to `libsys_airflow.dags.vendor.data_fetcher` where the DAG instance is created and scheduled. This will give the Vendor Loads web app the `dag_run_id` before it is run, so it will be able to update it.
2. Rename `VendorFile.expected_execution` to `VendorFile.dag_execution_date` in the vendor database, and make it a `DateTime` (not just a `Date`). Populate it from `libsys_airflow.dags.vendor.data_fetcher` when the DAG instance is created. *Question: what property should I use from the Dag instance?*
3. Update the Flask app to update the Dag instance when the user wants to reschedule the load, in addition to updating its own database.  *Question: Should I use the REST API or Python API for that?*

edsu:
It turns out adjusting the execution time of a DagRun is not something that most people do, and works against the grain of how Airtable is typically used. After some discussion we are going to try a different approach:

- [ ] Remove the dynamic creation of `default_data_processor` DAGs at the end of `data_fetcher`
- [ ] Create a new `load_scheduler` that runs hourly and queries the `vendor_files` table looking for files that need loading and creates `default_data_processor` DAGs as needed.

libsys_airflow/dags/vendor/data_fetcher.py
```diff
@@ -73,68 +70,27 @@ def setup():
...
     )
+
     archive_task(
         downloaded_files,
         params["download_path"],
         params["vendor_uuid"],
         params["vendor_interface_uuid"],
     )
 
-    dag_run_kwargs = generate_dag_run_kwargs.partial(
```
edsu, line 125: Generating the dag runs has been moved to the folio_load_scheduler DAG which runs on a schedule

libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -94,13 +94,6 @@ def setup():
             # Not yet supported in UI.
             params["archive_regex"] = processing_options.get("archive_regex")
 
-            vendor_file = VendorFile.load_with_vendor_interface(
-                vendor_interface, params["filename"], session
-            )
-            logger.info(f"vendor_file is {vendor_file}")
-            vendor_file.dag_run_id = context["run_id"]
-            session.commit()
-
```
edsu, line 103: Updating the VendorFile has been moved to the `folio_load_scheduler` DAG

libsys_airflow/dags/vendor/folio_load_scheduler.py
```diff
@@ -0,0 +1,80 @@
...
+                    }
+                )
+
+            session.commit()
+
+        return confs
+
+    dag_run_kwargs = generate_dag_run_kwargs()
+
+    TriggerDagRunOperator.partial(task_id="process_file").expand_kwargs(dag_run_kwargs)
```
edsu, line 80: This is a new DAG that runs hourly to see what files need to be loaded and starts a DagRun for them.

libsys_airflow/plugins/vendor/download.py
```diff
@@ -201,6 +209,7 @@ def _record_vendor_file(
             filesize=filesize,
             status=status,
             vendor_timestamp=vendor_timestamp,
+            expected_load_time=expected_load_time,
```
edsu, line 212: The expected_load_time can be set right away.

libsys_airflow/plugins/vendor/emails.py
```diff
@@ -48,13 +48,15 @@ def files_fetched_email_task(
...
-    )
+    email = os.getenv('VENDOR_LOADS_TO_EMAIL')
+    if email:
+        send_email(
+            email,
+            f"{vendor_name} ({vendor_code}) - {vendor_interface_uuid} - Daily Fetch Report ({date.today().isoformat()})",
+            _html_content(
+                vendor_name, vendor_code, vendor_interface_uuid, downloaded_files
+            ),
+        )
```
edsu, line 59: Guard against VENDOR_LOADS_TO_EMAIL not being set in the environment (not in dev compose environment).

libsys_airflow/plugins/vendor/models.py
```diff
@@ -158,7 +161,7 @@ class VendorFile(Model):
     filesize = Column(Integer, nullable=False)
     vendor_timestamp = Column(DateTime, nullable=True)
     loaded_timestamp = Column(DateTime, nullable=True)
-    expected_execution = Column(Date, nullable=True)
+    expected_load_time = Column(DateTime, nullable=True)
```
edsu, line 164: The expected_load_time is a datetime now, not just a date.

libsys_airflow/plugins/vendor/models.py
```diff
@@ -186,3 +189,21 @@ def load_with_vendor_interface(
...
+        """
+        return session.scalars(
+            select(VendorFile)
+            .filter(
+                VendorFile.status.in_([FileStatus.fetched]),
+            )
+            .filter(VendorFile.expected_load_time <= datetime.now(timezone.utc))
+            .limit(1000)
+            .order_by(VendorFile.created.asc())
+        ).all()
```
edsu, line 209: A class method for looking up VendorLoad files that need processing. It limits to the last 1000 files but maybe this will need to be reduced depending on how much Airflow can handle?

libsys_airflow/plugins/vendor_app/templates/vendors/file.html
```diff
@@ -0,0 +1,98 @@
...
+  </div>
+  <div class="form-group">
+    <div class="col-sm-1">
+      <button class="btn btn-success">Save</button>
+    </div>
+  </div>
+  {% endif %}
+</form>
+
+{% endblock %}
```
edsu, line 98: A view for a VendorFile that lets you edit the expected_load_time if it hasn't been loaded already.

libsys_airflow/plugins/vendor_app/templates/vendors/interface.html
```diff
@@ -97,6 +100,7 @@ <h2>Completed Loads</h2>
   </thead>
   {% for file in interface.processed_files %}
   <tr>
+    <td><a href="{{ url_for('VendorManagementView.file', file_id=file.id) }}">{{ file.id }}</a></td>
```
edsu, line 103: Link to the File view.

libsys_airflow/plugins/vendor_app/vendor_management.py
```diff
@@ -203,6 +203,21 @@ def _create_vendor_file(self, interface, file_upload, filepath, session):
...
+            try:
+                file.expected_load_time = datetime.fromisoformat(
+                    request.form['expected-load-time']
+                )
+                session.commit()
+            except ValueError:
+                flash("invalid date: {request.form['expected-load-time']}")
+
+        return self.render_template("vendors/file.html", file=file)
+
```
edsu, line 220: Server side piece to the VendorFile view. It also lets a POST update the expected_load_time.

tests/apps/vendor_app/test_file_view.py
```diff
@@ -0,0 +1,114 @@
...
+        assert expected_load_time
+        expected_load_time = expected_load_time.attrs['value']
+        expected_load_time = datetime.fromisoformat(expected_load_time)
+        assert expected_load_time.date() == tomorrow.date()
+        assert expected_load_time.time() == tomorrow.time().replace(microsecond=0)
+
+        # peek in the database to see if it was updated there
+        vendor_file = session.get(VendorFile, 1)
+        assert vendor_file.expected_load_time.date() == tomorrow.date()
+        assert vendor_file.expected_load_time.time() == tomorrow.time()
```
edsu, line 114: Test that the VendorFile view is working.

tests/vendor/test_schedule.py
```diff
@@ -0,0 +1,126 @@
...
+    return mock_hook
+
+
+def test_schedule(engine):
+    with Session(engine) as session:
+        vendor_files = VendorFile.ready_for_folio_load(session)
+        # only "fetched" fils with expected_load_time in the past should be returned
+        assert len(vendor_files) == 2
+        # results should be in ascending order by when they were fetched
+        assert vendor_files[0].created < vendor_files[1].created
```
edsu, line 126: Test that `VendorFile.ready_for_folio_load` works as expected.

vendor_loads_migration/versions/a634337e2c85_expected_load_time.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision = 'a634337e2c85'
+down_revision = 'bdc18b238b0c'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('vendor_files', sa.Column('expected_load_time', sa.DateTime(), nullable=True))
```
edsu, line 21: Rename the column to reflec that it's the expected load time, and make it a DateTime.

libsys_airflow/dags/vendor/folio_load_scheduler.py
```diff
@@ -0,0 +1,80 @@
...
+            # as a practical matter should we limit this further?
+            vendor_files = VendorFile.ready_for_folio_load(session)
+            if len(vendor_files) == 0:
+                return []
+
+            for vendor_file in vendor_files:
+                vendor_interface = vendor_file.vendor_interface
+                vendor = vendor_interface.vendor
+
+                execution_date = datetime.now()
```
jermnelson, line 47: Does it matter the execution_date uses the system's local timestamp vs. using `utcnow`?

libsys_airflow/plugins/vendor/download.py
```diff
@@ -191,8 +191,16 @@ def _record_vendor_file(
         existing_vendor_file = VendorFile.load_with_vendor_interface(
             vendor_interface, filename, session
         )
+
         if existing_vendor_file:
             session.delete(existing_vendor_file)
+
+        expected_load_time = datetime.now()
```
jermnelson, line 198: See question above about `now` vs. `utcnow`

libsys_airflow/plugins/vendor/models.py
jermnelson: Would setting the `max_active_runs` parameter for the processing DAGs address this problem?  Any DAG Runs beyond that threshold would be queued up, This would require any processing dags beside the default to also have the `max_active_runs` set as well.

bin/files
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
+
+# A utility for listing files that are ready to be loaded into Folio.
```
justinlittman, line 3: Can you say more about who might want to use this to solve what problem?

bin/files
```diff
@@ -0,0 +1,36 @@
...
+
+
+def main():
+    db_url = os.getenv(
+        'AIRFLOW_CONN_VENDOR_LOADS',
+        'postgresql+psycopg2://airflow:airflow@localhost:5432/vendor_loads',
+    )
+    engine = create_engine(db_url)
+    Session = sessionmaker(bind=engine)
+    session = Session()
```
justinlittman, line 25: This should probably be a `with Session() as session:`

jermnelson (approved):
A couple of questions related to execution timestamps and file limits but not blocking.

libsys_airflow/plugins/vendor/emails.py
justinlittman: You should be able to use `_email_enabled()` see L62.

libsys_airflow/plugins/vendor/models.py
justinlittman: 1K seems fine until proven otherwise.

libsys_airflow/plugins/vendor_app/templates/vendors/dashboard.html
```diff
@@ -24,6 +25,7 @@ <h1>Vendor Management</h1>
     <tbody>
       {% for file in in_progress_files %}
       <tr>
+        <td><a href="{{ url_for('VendorManagementView.file', file_id=file.id) }}">{{ file.id }}</a></td>
```
justinlittman, line 28: Instead of displaying the DB id, just link the filename?

libsys_airflow/plugins/vendor_app/templates/vendors/interface.html
```diff
@@ -72,21 +73,23 @@ <h2>Pending Loads</h2>
   </thead>
   {% for file in interface.pending_files %}
   <tr>
+    <td><a href="{{ url_for('VendorManagementView.file', file_id=file.id) }}">{{ file.id }}</a></td>
```
justinlittman, line 76: Same here re: db id.

justinlittman (changes requested):
Most of these are just suggestions for your consideration.

libsys_airflow/plugins/vendor/models.py
justinlittman: The main reason for this limit was to address Airflow's XCOM limitation of an array (must have <1024 entries).

libsys_airflow/plugins/vendor/models.py
```diff
@@ -186,3 +189,21 @@ def load_with_vendor_interface(
             .where(cls.vendor_interface_id == vendor_interface.id)
             .where(cls.vendor_filename == filename)
         ).first()
+
+    @classmethod
+    def ready_for_folio_load(cls, session: Session) -> List["VendorFile"]:
```
justinlittman, line 194: Can this be called "ready_for_data_import" for consistency?

libsys_airflow/plugins/vendor/models.py
```diff
@@ -186,3 +189,21 @@ def load_with_vendor_interface(
...
+        """
+        Returns a list of VendorFile objects that are ready for loading into
+        Folio. Thse are files that have a status of "fetched" and which have an
+        expected_load_time in the past. Results are ordered in ascending order
+        of when they were fetched, and no more than 1000 are returned at a time.
+        """
+        return session.scalars(
+            select(VendorFile)
+            .filter(
+                VendorFile.status.in_([FileStatus.fetched]),
```
justinlittman, line 204: This needs to account for SKIP IMPORT vendor interfaces. These are vendor interfaces in which the folio_data_import_profile_uuid is null.

One possibility is to set the expected_load_time to null for these earlier in dowload.py and then ignoring any in this query with null expected_load_time.

libsys_airflow/plugins/vendor/emails.py
edsu: Fixed!

libsys_airflow/dags/vendor/folio_load_scheduler.py
edsu: I think you're right. We should be using utcnow() here I think?

justinlittman:
Based on testing in stage, there is a mismatch between the TZ stored in the DB for expected_load_time and the ready for load query.

bin/files
edsu: Ok, FWIW I was copying the approach taken in `bin/seed_vendors`.

libsys_airflow/plugins/vendor_app/templates/vendors/dashboard.html
edsu: The filename is currently linked to the download.

libsys_airflow/plugins/vendor_app/templates/vendors/interface.html
edsu: The filename is currently linked to the download here too.

libsys_airflow/plugins/vendor/models.py
edsu: I deemed `data_import` to be a bit too vague and wanted to be more explicit.

justinlittman: Agree, but `data_import` is the name of the Folio module/API.

libsys_airflow/plugins/vendor_app/templates/vendors/file.html
```diff
@@ -0,0 +1,98 @@
...
+  <div class="form-group">
+    <label for="updated" class="col-sm-2 col-form-label">Updated:</label>
+    <div class="col-sm-8" id="updated">
+      {{ file.updated }}
+    </div>
+  </div>
+  <div class="form-group">
+    <label for="vendor-filename" class="col-sm-2 col-form-label">File Name:</label>
+    <div class="col-sm-8" id="vendor-filename">
+      {{ _macros.fileDownload(file, file.vendor_filename) }}
```
justinlittman, line 54: For consistency, can you use the download icon?

edsu: I was being consistent with the Dashboard, but have brought both in line with the Interface view now.

edsu:
I'm going to ticket the timezone problem separately from this PR since there are already too many changes bundled into this PR.

vendor_loads_migration/versions/1d954292be5e_merge_expected_load_time_changes.py
```diff
@@ -0,0 +1,24 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    pass
+
+
+def downgrade() -> None:
+    pass
```
edsu, line 24: This came in because of a dual-heads migration conflict.

docker-compose.yaml
```diff
@@ -79,10 +79,9 @@ x-airflow-common:
     AIRFLOW_VAR_FOLIO_USER: ${AIRFLOW_VAR_FOLIO_USER}
     AIRFLOW_VAR_FOLIO_PASSWORD: ${AIRFLOW_VAR_FOLIO_PASSWORD}
     AIRFLOW_VAR_OKAPI_URL: ${AIRFLOW_VAR_OKAPI_URL}
-    AIRFLOW_VAR_DOWNLOAD_DAYS_AGO: 10
+    AIRFLOW_VAR_DOWNLOAD_DAYS_AGO: 10000
```
justinlittman, line 82: This should not change.

docker-compose.yaml
```diff
@@ -79,10 +79,9 @@ x-airflow-common:
...
     AIRFLOW_CONN_VENDOR_LOADS: postgresql+psycopg2://airflow:airflow@postgres/vendor_loads
     SQLALCHEMY_SILENCE_UBER_WARNING: 1
-    SQLALCHEMY_WARN_20: 1
```
justinlittman, line 85: Why was this removed?

justinlittman (approved):
Approve pending docker-compose fixes.

docker-compose.yaml
edsu: Whoops, nice catch -- that was from me testing.

docker-compose.yaml
edsu: It was an accident, I will restore.

PR #502, opened 2023-05-25
Refactors expected_execution to archive_date and changes when it is r…

### LINKED ISSUE #501: Refactor "expected execution"
The VendorFile.expected_execution field is confusing. It is a date for execution, but it isn't clear what is being executed. And it is not "expected".

The intended purpose of this field is to support file archiving. This should be refactored so that its meaning and use is clearer.

justinlittman:
…ecorded in vendor file.

closes #501


libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -73,11 +76,9 @@ def setup():
             params["add_fields"] = processing_options.get("add_fields")
             params["archive_regex"] = processing_options.get("archive_regex")
 
-            vendor_file = session.scalars(
-                select(VendorFile)
-                .where(VendorFile.vendor_filename == params["filename"])
-                .where(VendorFile.vendor_interface_id == vendor_interface.id)
-            ).first()
+            vendor_file = load_vendor_file_with_vendor_interface(
```
justinlittman, line 79: Added a helper for loading VendorFiles. Used here and multiple locations below.

libsys_airflow/plugins/vendor/archive.py
```diff
@@ -1,27 +1,68 @@
...
+    filename: str,
+    vendor_file: VendorFile,
+    session: Session,
+):
+    os.makedirs(archive_path, exist_ok=True)
+
+    download_filepath = _filepath(download_path, filename)
+    archive_filepath = _filepath(archive_path, filename)
+    shutil.copyfile(download_filepath, archive_filepath)
+    vendor_file.archive_date = date.today()
```
justinlittman, line 63: Archive task now updates the archive date.

libsys_airflow/plugins/vendor/download.py
```diff
@@ -70,7 +74,6 @@ def ftp_download_task(
     download_path: str,
     filename_regex: str,
     vendor_interface_uuid: str,
-    expected_execution: str,
```
justinlittman, line 73: expected_execution no longer needs to be dragged through download code.

libsys_airflow/plugins/vendor/models.py
```diff
@@ -147,7 +147,8 @@ class VendorFile(Model):
     filesize = Column(Integer, nullable=False)
     vendor_timestamp = Column(DateTime, nullable=True)
     loaded_timestamp = Column(DateTime, nullable=True)
-    expected_execution = Column(Date, nullable=False)
+    expected_execution = Column(Date, nullable=True)
```
justinlittman, line 150: Ed will separately be re-using expected_execution so leaving in model.

libsys_airflow/plugins/vendor_app/vendors.py
```diff
@@ -124,20 +127,39 @@ def file_upload(self, interface_id):
 
     def _handle_file_upload(self, interface_id, file_upload):
         session = Session()
-        execution_date = datetime.now()
         interface = session.query(VendorInterface).get(interface_id)
-        filepath = self._save_file(
+        _download_path = download_path(
```
justinlittman, line 131: This is just refactoring to get the VendorFile created before archiving so that the archive_date can be set on it.

libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -73,11 +74,9 @@ def setup():
...
```
edsu, line 77: Does it make sense to add this helper function as a class method on VendorInterface?

edsu (approved):
Thanks for clarifying this @justinlittman!

libsys_airflow/dags/vendor/default_data_processor.py
justinlittman: It does. Thanks for the suggestion. Changes made.

PR #504, opened 2023-05-26
Skip data import when no data import profile.

### LINKED ISSUE #496: Update the default processing DAG to treat null folio_data_import_profile_uuid as a "fetch only" interface
For order records, where the leheigh tool is expected to be used initially for data ingestion, we still need to automatically fetch the files for the configured vendor interface and notify users that files have been fetched (and display in the UI) but NOT execute a data import.

For this ticket, update the default processing dag to check if folio_data_import_profile_uuid is null before the data_import task. If it's null end the dag successfully. Otherwise proceed to data import.

A follow up ticket will specify the notification process.

justinlittman:
closes #496

image

### DIFF
--- a/libsys_airflow/dags/vendor/default_data_processor.py
+++ b/libsys_airflow/dags/vendor/default_data_processor.py
@@ -10,7 +10,10 @@ from airflow.providers.postgres.hooks.postgres import PostgresHook
 
 from libsys_airflow.plugins.vendor.marc import process_marc_task, batch_task
 from libsys_airflow.plugins.vendor.paths import download_path
-from libsys_airflow.plugins.folio.data_import import data_import_task
+from libsys_airflow.plugins.folio.data_import import (
+    data_import_task,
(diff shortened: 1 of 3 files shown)


libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -99,8 +102,10 @@ def setup():
         params["add_fields"],
     )
     batch_filenames = batch_task(params["download_path"], filename)
+    data_import_branch = data_import_branch_task(params["dataload_profile_uuid"])
     data_import = data_import_task(
         params["download_path"], batch_filenames, params["dataload_profile_uuid"]
     )
 
     process_marc >> batch_filenames
+    data_import_branch >> data_import
```
edsu, line 111: It's still hard for me to follow the TaskFlow style and the `>>` style of connecting tasks.

edsu (approved):

PR #507, opened 2023-05-26
Set DB connection pool max size.

justinlittman:


edsu (approved):

docker-compose.prod.yaml
```diff
@@ -58,6 +58,7 @@ x-airflow-common:
     AIRFLOW__CORE__EXECUTOR: CeleryExecutor
     # For backward compatibility, with Airflow <2.3
     AIRFLOW__CORE__SQL_ALCHEMY_CONN: postgresql+psycopg2://${DATABASE_USERNAME}:${DATABASE_PASSWORD}@${DATABASE_HOSTNAME}/libsys
+    AIRFLOW__DATABASE__SQL_ALCHEMY_MAX_OVERFLOW: ${DATABASE_MAX_OVERFLOW:-25}
```
edsu, line 61: Is `-25` the same as `-1` here? Is it worth documenting that this means no limit to the number of extra connections that will be opened?

edsu (changes requested):

docker-compose.prod.yaml
justinlittman: `:-` is docker compose syntax for "use the default unless provided". In this case, the default is 25.

edsu (approved):

PR #514, opened 2023-05-26
Add Folio job execution ID to model and display links thereto in dashboard

### LINKED ISSUE #508: Link to Folio data import on vendor management dashboard
The vendor management dashboard, as implemented in #503, satisfies most of what was written up in #473---but not all. One of the missing pieces is to link files in the `loading_error` state in the `Errors` table to the associated Folio data import so that users can get more context about the loading error. See https://github.com/sul-dlss/libsys-airflow/issues/473 for the design.

The links would look like `https://folio-test.stanford.edu/data-import/job-summary/d7460945-6f0c-4e74-86c9-34a8438d652e`

mjgiarlo:
Fixes #508

Includes:
* Add Folio job execution UUID column to VendorFile model
  * This commit restores to the VendorFile model a column that was previously removed from the DataLoad model. When the data processor runs and gets the Folio job execution ID, that value is saved to all associated VendorFile instances. This is needed so we can construct links to Folio job executions, e.g., so users can get more information about vendor file loading errors (#508).
* Display links to Folio job executions in vendor management dashboard
  * This commit includes a change to the data import task that makes it possible to associate uploaded files in the vendor management app with files uploaded to Folio, which are always chunked into at least one chunk, which has the effect of adding a `_1` infix to the uploaded filename just before the file extension. So, uploading `data.mrc` to the vendor management app winds up loading `data_1.mrc` in Folio-land. To successfully associate these infixed files with the filenames in the `VendorFile` table, this commit maps over the `batch_filenames` in the data import task, dropping the `_n` infix and uniq'ing the list of filenames, which allows the filenames from Folio-land to match those in the vendor_loads database.


libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -103,5 +104,12 @@ def setup():
         params["download_path"], batch_filenames, params["dataload_profile_uuid"]
     )
 
+    VendorFile.update_batch_with_job_execution_id(
```
justinlittman, line 107: This should probably be in data import. I don't think it will work correctly in this location.

libsys_airflow/plugins/vendor/models.py
```diff
@@ -185,3 +186,16 @@ def load_with_vendor_interface(
             .where(cls.vendor_interface_id == vendor_interface.id)
             .where(cls.vendor_filename == filename)
         ).first()
+
+    @classmethod
+    def update_batch_with_job_execution_id(
```
justinlittman, line 191: Since this is only going to be used in data import, I would put it there.

justinlittman (changes requested):

libsys_airflow/dags/vendor/default_data_processor.py
mjgiarlo: OK, moved this functionality into `data_import.py`

libsys_airflow/plugins/vendor/models.py
mjgiarlo: K, I moved it there. Thanks!

libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -100,7 +100,10 @@ def setup():
     batch_filenames = batch_task(params["download_path"], filename)
     data_import_branch = data_import_branch_task(params["dataload_profile_uuid"])
     data_import = data_import_task(
-        params["download_path"], batch_filenames, params["dataload_profile_uuid"]
+        params["download_path"],
+        batch_filenames,
+        params["dataload_profile_uuid"],
+        params["vendor_interface_uuid"],
```
mjgiarlo, line 106: We need to know the vendor interface in the data import task now.

libsys_airflow/plugins/folio/data_import.py
```diff
@@ -45,16 +46,26 @@ def data_import_branch_task(dataload_profile_uuid: Optional[str]):
...
 
 
 def data_import(
-    download_path, batch_filenames, dataload_profile_uuid, folio_client=None
+    download_path,
+    batch_filenames,
+    dataload_profile_uuid,
+    vendor_interface_uuid,
+    folio_client=None,
+    session=None,
```
mjgiarlo, line 68: Allow injection of a session for testing purposes.

libsys_airflow/plugins/vendor_app/templates/vendors/_macros.html
```diff
@@ -1,3 +1,11 @@
 {% macro fileDownload(file, link_text='<span class="glyphicon glyphicon-save" aria-hidden="true" style="color: #017cee;"></span>') -%}
   {% if file.status != 'purged' %}<a href="{{ url_for('VendorManagementView.download_file', file_id=file.id) }}">{{ link_text|safe }}</a>{% endif %}
 {%- endmacro %}
+
+{% macro renderFileStatus(file, folio_base_url) -%}
+  {% if file.folio_job_execution_uuid == None %}
```
mjgiarlo, line 6: NOTE: Yes, we're adding links not only to loading_error files in the dashboard but any files that have a job execution ID. More linkage = better?

libsys_airflow/plugins/vendor_app/vendors.py
```diff
@@ -44,7 +45,7 @@ def dashboard(self):
         )
         errors_files = (
             Session()
-            .query(Vendor)
+            .query(VendorFile)
```
mjgiarlo, line 48: This was a bug/typo.

libsys_airflow/plugins/folio/data_import.py
edsu: Can the session be mocked in the unit test?

libsys_airflow/plugins/folio/data_import.py
```diff
@@ -79,11 +90,26 @@ def data_import(
         upload_definition,
         data_type,
     )
+    with session or _session() as db_session:
```
edsu, line 93: If session is mocked correctly this should not be needed.

libsys_airflow/plugins/vendor_app/templates/vendors/_macros.html
edsu: I don't know if this changes things but I've been working on a "file" view as part of #480

edsu (dismissed):

libsys_airflow/plugins/folio/data_import.py
justinlittman: Agree with Ed -- we've gotten away with not having to inject sessions in other tests. It would be slightly preferable to follow this pattern.

libsys_airflow/plugins/folio/data_import.py
```diff
@@ -79,11 +90,26 @@ def data_import(
...
+        for filename in _unique_dechunked_filenames(batch_filenames):
```
justinlittman, line 94: I think that there will only be 1 job execution id that needs to be set for 1 VendorFile. Further, you get the VendorFile name by including it as an argument for the task (i.e., no need to dechunk the filenames).

justinlittman (changes requested):

libsys_airflow/plugins/folio/data_import.py
mjgiarlo: @edsu @justinlittman Good call, you bet.

libsys_airflow/plugins/folio/data_import.py
mjgiarlo: 🙌🏻

libsys_airflow/plugins/folio/data_import.py
mjgiarlo: Done.

libsys_airflow/plugins/folio/data_import.py
mjgiarlo: Removed.

libsys_airflow/plugins/folio/data_import.py
mjgiarlo: I think I understood that and made the change. Have another look?

justinlittman (approved):
Approve pending green.

PR #515, opened 2023-05-26
Max overflow configuration fix.

justinlittman:
3rd time is a charm?

### DIFF
--- a/libsys_airflow/plugins/vendor_app/database.py
+++ b/libsys_airflow/plugins/vendor_app/database.py
@@ -14,7 +14,7 @@ class EngineWrapper:
     def __getattr__(self, name):
         if self.engine is None:
             self.engine = PostgresHook("vendor_loads").get_sqlalchemy_engine(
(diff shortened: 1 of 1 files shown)


libsys_airflow/plugins/vendor_app/database.py
```diff
@@ -14,7 +14,7 @@ def __init__(self):
     def __getattr__(self, name):
         if self.engine is None:
             self.engine = PostgresHook("vendor_loads").get_sqlalchemy_engine(
-                max_overflow=int(os.getenv("DATABASE_MAX_OVERFLOW", "10"))
+                {"max_overflow": int(os.getenv("DATABASE_MAX_OVERFLOW", "10"))}
```
edsu, line 17: Should the default be 20 to match what's in the Docker config?

edsu (approved):

libsys_airflow/plugins/vendor_app/database.py
justinlittman: It could, but they're all really guesses.

PR #529, opened 2023-05-31
Seed all vendors

### LINKED ISSUE #520: Update database seeding and refreshing to include all vendors
Remove the filtering of vendors to our day 1 vendors so that stage and prod can be deployed with all vendors. Also remove any filtering from the refresh task.

lwrubel:
Resolves #520 to seed database with all vendors or a subset.

### DIFF
--- a/bin/seed_vendors
+++ b/bin/seed_vendors
@@ -44,21 +44,25 @@ class VendorLoader:
         """
         Returns vendors from FOLIO
         """
-        vendors = [
-            "AMALIV-SUL",
-            "CASALI-SUL",
-            "COUTTS-SUL",
-            "HARRAS-SUL",
-            "SFX",
-            "YANKEE-SUL",
-        ]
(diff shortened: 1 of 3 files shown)


bin/seed_vendors
```diff
@@ -91,7 +95,12 @@ def main(opts):
         os.getenv('AIRFLOW_VAR_FOLIO_USER'),
         os.getenv('AIRFLOW_VAR_FOLIO_PASSWORD'),
     )
-    for organization in vendor_load.vendors["organizations"]:
+    acq_uuids = set([org["acqUnitIds"][0] for org in vendor_load.vendors ])
+    acq_data = {}
+    for uuid in acq_uuids:
+        acq_data[uuid] = vendor_load.get_acquisitions_unit_name(uuid)
```
lwrubel, line 101: There are only a handful of acquisitions units, so look up this set of names in advance, rather than with every single vendor.

aaron-collier (approved):

PR #546, opened 2023-06-01
Adds support for Honeybadger.

### LINKED ISSUE #545: Add Honeybadger to Vendor Management

justinlittman:
closes #545


docker-compose.prod.yaml
```diff
@@ -92,6 +92,8 @@ x-airflow-common:
     _PIP_ADDITIONAL_REQUIREMENTS: ${_PIP_ADDITIONAL_REQUIREMENTS:-}
     DATABASE_USERNAME: ${DATABASE_USERNAME}
     DATABASE_MAX_OVERFLOW: ${DATABASE_MAX_OVERFLOW:-20}
+    HONEYBADGER_ENVIRONMENT: ${HONEYBADGER_ENVIRONMENT}
```
mjgiarlo, line 95: Does HB use this variable or does it need to be `HONEYBADGER_ENV` (https://docs.honeybadger.io/lib/ruby/getting-started/environments/)?

pyproject.toml
```diff
@@ -21,6 +21,8 @@ python-dotenv = "^1.0.0"
 alembic = "^1.10.4"
 psycopg2-binary = "^2.9.6"
 python-magic = "^0.4.27"
+honeybadger = "^0.15.2"
+blinker = "^1.6.2"
```
mjgiarlo, line 25: What's blinker?

jmartin-sul: looks like a library for sending/receiving messages/signals/events?  https://pypi.org/project/blinker/ 💬 

> Blinker provides a fast dispatching system that allows any number of interested parties to subscribe to events, or “signals”.

already seems to be an indirect dependency via something else?  no diff on the `poetry.lock` lines where it's included: https://github.com/sul-dlss/libsys-airflow/pull/546/files#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89R731-R733

i don't see it referenced in this PR outside of `pyproject.toml` or `poetry.lock`, so maybe it's cruft from an earlier iteration of this work?

justinlittman: The HB docs said it needed to be added for Flask integration so I explicitly added it.

mjgiarlo: @justinlittman OK, cool. What would you think of adding this as a comment?

PR #553, opened 2023-06-02
Make iteration deterministic for test

jermnelson:
With the switch to the Ubuntu image for CI workflows revealed difference in iteration with Mac image.  PR makes sorting for the `migration/iterations` directories deterministic.


edsu (approved):
Nice, thanks for investigating!

PR #558, opened 2023-06-02
[HOLD] DAG to sync vendor data from folio and deprecate removed interfaces

lwrubel:
Requires tests and more testing, as well as confirming UI makes sense for deprecated interfaces.


libsys_airflow/plugins/vendor/sync.py
```diff
@@ -0,0 +1,147 @@
...
+@task()
+def sync_data_task():
+    folio_client = _folio_client()
+    sync_data(folio_client)
+
+
+def sync_data(folio_client):
+    organizations = folio_client.organizations
+    acq_names = _get_acquisitions_unit_names(
+        set([org["acqUnitIds"][0] for org in organizations]), folio_client
```
jermnelson, line 63: Unrelated to this PR, I was seeing an IndexError in similar code when an organization did not have an assigned Acquisitions Unit Id with the `org["acqUnitIds"]` returned an empty list. This may not be issue if all of the organizations have at least one acquisition unit. You could add a guard clause to this list comprehension like 
```
set([org["acqUnitIds"][0] for org in organizations if len(org["acqUnitIds"]) > 0)

**Review by jermnelson — CHANGES_REQUESTED** [review]
Suggestion for a small update to the list comprehension to guard against an empty acqUnitIds list.

File: libsys_airflow/plugins/vendor/sync.py [hunk 2]
```diff
@@ -0,0 +1,147 @@
...
+            # Deprecate removed interfaces
+            existing_vendor_interfaces = session.scalars(
+                select(VendorInterface.folio_interface_uuid).where(
+                    VendorInterface.vendor_id == vendor.id
+                )
+            ).all()
+            removed = list(
+                set(existing_vendor_interfaces).difference(organization["interfaces"])
+            )
+            for interface_uuid in removed:
```
justinlittman, line 135: Interfaces which are upload only (i.e., have a null folio_interface_uuid) should not be removed. I believe with this logic they are.

justinlittman (changes requested):

libsys_airflow/plugins/vendor/sync.py
lwrubel: I see in the models that:
`acquisitions_unit_from_folio = Column(String(36), unique=False, nullable=False)`

Should this field be nullable?

lwrubel:
Hoping it's less confusing, I flipped the `removed_in_folio` field to a positive and renamed it `assigned_in_folio`. 

Screenshot 2023-06-06 at 10 20 56 AM

Unassign/assign is terminology used in the FOLIO UI. Upload-only interfaces will also have `assigned_in_folio` be False.

@justinlittman I added logic to avoid changing the status of upload-only interfaces when the sync DAG runs.

libsys_airflow/plugins/vendor/sync.py
lwrubel: Thanks for catching. Updated.

lwrubel: Also added a test case to cover sync on upload-only.

libsys_airflow/plugins/vendor/sync.py
```diff
@@ -0,0 +1,155 @@
...
+                select(VendorInterface.folio_interface_uuid).where(
+                    VendorInterface.vendor_id == vendor.id
+                )
+            ).all()
+            unassigned = list(
+                set(existing_vendor_interfaces).difference(organization["interfaces"])
+            )
+            for interface_uuid in unassigned:
+                if interface_uuid is not None:  # do not unassign upload-only interfaces
+                    unassigned_interface = (
```
lwrubel, line 144: Fixing logic to keep upload-interfaces from getting reset.

libsys_airflow/plugins/vendor/sync.py
jermnelson: Since acquisitions Unit Ids are not in the required fields for [Organizations ](https://github.com/folio-org/acq-models/blob/505ff2e91f7d5f0e1fc164ad959d4af7e8d76900/mod-orgs/schemas/organization.json#L430-L433 ), I think this field should be nullable. Good catch1

lwrubel: I've added a migration to make `acquisitions_unit_from_folio` nullable.

justinlittman (approved):
@lwrubel There have been some recent db migrations added. Does this need to be adjusted?

PR #562, opened 2023-06-02
Adds ability to manually invoke a fetch for an interface.

justinlittman:


libsys_airflow/plugins/vendor_app/vendor_management.py
```diff
@@ -256,3 +269,20 @@ def _trigger_processing_dag(self, file):
             },
         )
         logger.info(f"Triggered DAG {dag} for {file.vendor_filename}")
+
+    def _trigger_fetcher_dag(self, interface):
```
edsu, line 273: Why the `_` ?

justinlittman: It is intended to be private/helper method (and is consistent with other methods in the class).

edsu: If it is being called from outside he class definition I think it's no longer really considered private.

ahafele:
If a data import profile is selected for this interface, will this button will fetch and send to data import? If no data import profile is selected, will this button just fetch? 
If above two questions are yes, could the button say "Process now"?

justinlittman:
Yes, yes, and yes with the caveat that there may be up to an hour wait between fetching and processing.

justinlittman:
image

aaron-collier (approved):

PR #585, opened 2023-06-06
Use correct env variable from HB env.

### LINKED ISSUE #577: HB reporting all environments as production environment
image

The `HONEYBADGER_ENV` environment variable is being set in docker image. I have a feeling for python, it might be `HONEYBADGER_ENVIRONMENT`.

justinlittman:
closes #577


docker-compose.prod.yaml
```diff
@@ -92,6 +92,9 @@ x-airflow-common:
     _PIP_ADDITIONAL_REQUIREMENTS: ${_PIP_ADDITIONAL_REQUIREMENTS:-}
     DATABASE_USERNAME: ${DATABASE_USERNAME}
     DATABASE_MAX_OVERFLOW: ${DATABASE_MAX_OVERFLOW:-20}
+    # HB Python library uses HONEYBADGER_ENVIRONMENT.
+    HONEYBADGER_ENVIRONMENT: ${HONEYBADGER_ENV}
+    # HB Ruby library use HONEYBADGER_ENV.
```
justinlittman, line 97: Note that will be using HB Ruby library for deploy tracking.

PR #588, opened 2023-06-06
data_import: _record_status sets vendor_file.loaded_timestamp when updating .status to FileStatus.loaded

### LINKED ISSUE #555: loaded_timestamp not being set
image

It should be set in the `on_success_callback`: https://github.com/sul-dlss/libsys-airflow/blob/main/libsys_airflow/plugins/folio/data_import.py#L25-L26

jmartin-sul:
closes #555

i haven't done a data load myself yet, so i was a bit unsure how to test this realistically.  but happy to do that either locally or in a deployed env if there or instructions or if someone wouldn't mind pairing on it?

### DIFF
--- a/libsys_airflow/plugins/folio/data_import.py
+++ b/libsys_airflow/plugins/folio/data_import.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 import logging
 import os
 import pathlib
@@ -151,7 +152,7 @@ def _process_files(
     )
 
 
-def _record_status(context, status):
+def _record_status(context, status: FileStatus):
     vendor_interface_uuid = context["params"]["vendor_interface_uuid"]
     filename = context["params"]["filename"]
 
@@ -161,6 +162,8 @@ def _record_status(context, status):
     with Session(pg_hook.get_sqlalchemy_engine()) as session:
         vendor_file = VendorFile.load(vendor_interface_uuid, filename, session)
         vendor_file.status = status
+        if status is FileStatus.loaded:
+            vendor_file.loaded_timestamp = datetime.utcnow()
         session.commit()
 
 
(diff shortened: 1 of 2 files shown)


libsys_airflow/plugins/folio/data_import.py
```diff
@@ -161,6 +162,8 @@ def _record_status(context, status):
     with Session(pg_hook.get_sqlalchemy_engine()) as session:
         vendor_file = VendorFile.load(vendor_interface_uuid, filename, session)
         vendor_file.status = status
+        if status is FileStatus.loaded:
+            vendor_file.loaded_timestamp = datetime.utcnow()
```
jmartin-sul, line 166: on the one hand, it felt a bit weird adding this status-specific conditional logic to a general status update function.  on the other hand, it is private to the module, and doing something more generalizable, like having it take an optional callback function, seemed like overkill.  and the function is very much about `VendorFile.status`, which this timestamp is directly related to (and it's not like the function now knows about another class or variable).

happy to revisit if desired.

tests/test_data_import.py
```diff
@@ -248,16 +248,24 @@ def test_record_loading(context, pg_hook):
             select(VendorFile).where(VendorFile.vendor_filename == "3820230411.mrc")
         ).first()
         assert vendor_file.status == FileStatus.loading
+        assert vendor_file.loaded_timestamp is None
 
 
 def test_record_loaded(context, pg_hook):
+    before_record_loaded_time = datetime.utcnow()
     record_loaded(context)
+    after_record_loaded_time = datetime.utcnow()
```
jmartin-sul, line 257: this pattern for timestamp checking always feels a bit clunky to me (in any language), but i like the precision of it compared to just checking for a non-`None` timestamp.  happy to revisit if there's a more concise or idiomatic way to do this.

justinlittman: I'm not a huge fan of, as it seems to be a risk of being flaky. How about mocking out datetime? (Happy to pair on this.)

justinlittman (changes requested):

justinlittman (approved):

PR #617, opened 2023-06-09
Ensure MARC Records Encoded as UTF-8

### LINKED ISSUE #491: Handle wrongly encoded MARC Records
Our current strategy of reading [MARC records](https://github.com/sul-dlss/libsys-airflow/blob/f1d33d2c748614523e4f13379eb07821381f1dac/libsys_airflow/plugins/vendor/marc.py#L101-L104) from Vendors is to explicitly set the `to_unicode` param to True, the problem is that we have MARC records that encoded as MARC8 but have a value `a` in position 9 in the leader that indicate that the record is UTF-8 This results in records with MARC8 characters that are incorrectly encoded as UTF-8. 

For example, this a record contains Korean MARC8 characters in the 245 as read with our current code:

```
=245  10$a$1oWtoIoo].oL}oPJo\d(B
```

It should look like:
```
=245  10$a중국현대문학
```

After discussion with @ahafele and @shelleydoljack at LibSys Standup, we should follow the current Symphony workflow when using MarcEdit, files are manually processed by reading as MARC8 and then saved as UTF-8 by staff.

jermnelson:
Fixes #491


pyproject.toml
```diff
@@ -24,6 +23,7 @@ python-magic = "^0.4.27"
 honeybadger = "^0.15.2"
 # Needed according to HB docs.
 blinker = "^1.6.2"
+folio-migration-tools = {git = "https://github.com/FOLIO-FSE/folio_migration_tools.git"}
```
jermnelson, line 26: When the `folio-migration-tools` release a new version support pymarc 5.1.0, this should be pinned to that version or above instead of using the main branch for this dependency.

aaron-collier (approved):

Issue #613, opened 2023-06-09
Refactor fetch_only logic to include marc processing


aaron-collier:
We have determined that the use of fetch_only interfaces, especially when testing data import profile work, need to have applied the marc processing.

Update the logic around fetch_only interfaces to either go through the processing dag up to the point of data import, or a marc_processing only dag. 

The result should still be an email to the users that a file was processed "but not loaded".

justinlittman:
@aaron-collier The email work needs to be done. However, if I'm reading the code correctly the fetch_only workflow is already implemented as described. Do you have evidence to the contrary?

PR #620, opened 2023-06-12
Sends email when file processed, but not loaded.

justinlittman:
refs #613


edsu (approved):
I just want to register that (to me) "load" sounds more descriptive of what is happening than "processing". But I realize you are normalizing language with the name of the DAG.

justinlittman:
The primary motivation for the name change is because the default data processor DAG does not necessarily perform a load/data import. (The existing name made sense when a load was always performed.)

PR #635, opened 2023-06-12
Use upload definition state to determine file processing is complete

mjgiarlo:
Fixes sul-dlss/FOLIO-Project-Stanford#456

This commit changes how the file loaded sensor task determines whether a file has finished processing.

The prior implementation used the metadata provider API to get information about the job execution ID, and relied on a count of results from the job execution. We now understand this to have been flawed: if a job is processing multiple files, the job execution results will update after each file is successfully processed, which means the file sensor task is declaring victory too early. This leads to under-reporting file load information.

The implementation in this commit uses the upload definition within the data import API to track whether a given set of files has finished being processed by doing a `GET` on the upload definition and inspecting the response's `status` property. The values in this property can be `NEW`, `IN_PROGRESS`, `LOADED`, `COMPLETED`, or `ERROR` ([relevant API documentation](https://s3.amazonaws.com/foliodocs/api/mod-data-import/r/dataImport.html#data_import_uploaddefinitions__uploaddefinitionid__get)). The last two of these indicate processing is finished.

Since we are no longer using the job execution results to determine if processing is complete, it no longer made sense to include job results (SRS stats and instance stats) in the file loaded sensor, so that work is broken out into a small new job summary task.


libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -18,7 +18,8 @@
     file_not_loaded_email_task,
 )
 from libsys_airflow.plugins.vendor.extract import extract_task
-from libsys_airflow.plugins.vendor.file_load_report import report_when_file_loaded_task
+from libsys_airflow.plugins.vendor.file_loaded_sensor import file_loaded_sensor_task
```
mjgiarlo, line 21: As part of this work, rename the sensor task to remove the word `report` and add the word `sensor`, revealing intention more clearly.

libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -129,18 +130,22 @@ def setup():
         filename,
     )
 
-    file_loaded_sensor = report_when_file_loaded_task(
-        params["vendor_interface_uuid"], filename
+    file_loaded_sensor = file_loaded_sensor_task(
+        params["vendor_interface_uuid"], filename, data_import["upload_definition_id"]
     )
 
+    job_summary = job_summary_task(data_import["job_execution_id"])
```
mjgiarlo, line 137: Whereas the job summary task only relies on the job execution ID now.

libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -129,18 +130,22 @@ def setup():
...
+
     file_loaded_email_task(
         params["vendor_code"],
         params["vendor_interface_name"],
-        file_loaded_sensor,
+        data_import["job_execution_id"],
```
mjgiarlo, line 142: This task no longer directly relies on the file sensor. It relies on the job summary and the data import, and the job summary is what now relies on the file sensor.

libsys_airflow/plugins/vendor/emails.py
```diff
@@ -107,18 +106,21 @@ def _vendor_interface_url(vendor_interface_uuid):
...
     load_time: datetime,
     records_count: int,
+    srs_stats: dict,
+    instance_stats: dict,
 ):
     if not _email_enabled():
         logger.info("Email not enabled.")
         return
 
-    file_loaded_info = json.loads(file_loaded_json)
```
mjgiarlo, line 120: We can get out of the kludgy business of stuffing JSON strings in xcom now.

libsys_airflow/plugins/vendor/file_loaded_sensor.py
```diff
@@ -0,0 +1,61 @@
+import logging
```
mjgiarlo, line 1: This file was renamed for clarity.

libsys_airflow/plugins/vendor/file_loaded_sensor.py
```diff
@@ -0,0 +1,61 @@
...
+            return PokeReturnValue(is_done=False)
+
+    upload_definition = folio_client.get(
+        f"/data-import/uploadDefinitions/{upload_definition_id}"
+    )
+
+    if upload_definition["status"] in ["COMPLETED", "ERROR"]:
+        return PokeReturnValue(is_done=True)
+
+    return PokeReturnValue(is_done=False)
```
mjgiarlo, line 61: The sensor task is now greatly simplified now that we no longer have to hit the job summary endpoint and count up result stats, and as also need to account for the "endpoint URL is 404ing" since the upload definition endpoint is already up and running (there's a slight delay before the job summary endpoint is created).

libsys_airflow/plugins/vendor/job_summary.py
```diff
@@ -0,0 +1,30 @@
...
+def job_summary(
+    job_execution_id: str,
+    client=None,
+) -> dict:
+    folio_client = client or _folio_client()
+    summary = folio_client.get(f"/metadata-provider/jobSummary/{job_execution_id}")
+    return {
+        "srs_stats": summary.get("sourceRecordSummary", {}),
+        "instance_stats": summary.get("instanceSummary", {}),
+    }
```
mjgiarlo, line 30: As a result of disentangling the sensor from the stats, the stats portion is now trivially simple.

justinlittman (approved):
This feels much cleaner.

PR #640, opened 2023-06-13
Set VendorFile status to "loaded"

### LINKED ISSUE #497: Add the ability for users to set the status of a file in the UI
For fetch only interfaces, or situations where a user may be manually handling a problematic file, they will need the ability to mark a file as processed in the interface to avoid duplication or confusion. 

For the purposes of this ticket, it's likely sufficient to add a "processed" button that sets the status on the file and not be concerned with being able to set the other statuses, that are mostly data flow and error statuses.

edsu:
Add the ability for the user to set the status to "loaded" for VendorFiles that have not already been loaded or purged.

The select box only allows transitioning from the current status to "loaded". Once it has been marked loaded (or it has been otherwise purged) the select box does not display and the status is displayed read-only.

I went with setting `VendorFile.loaded_timestamp` to the current time when it is marked loaded.

https://github.com/sul-dlss/libsys-airflow/assets/33829/4dbd89c6-d728-42a9-ac41-ce8177719b0f

Resolves #497

### DIFF
--- a/libsys_airflow/plugins/vendor/models.py
+++ b/libsys_airflow/plugins/vendor/models.py
@@ -176,6 +176,10 @@ class FileStatus(enum.Enum):
     purged = "purged"
     skipped = "skipped"  # Files on the FTP server that are not fetched because they are before the download window.
 
+    def can_set_loaded(self):
+        """Return a boolean indicating whether the FileStatus can be manually set to loaded by a user"""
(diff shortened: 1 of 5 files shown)


lwrubel:
Once these are set to "loaded" they show up in the "Sent to Data Import" table at the bottom, where loaded files are listed. Will the Reload action work on these files manually set to loaded?

PR #650, opened 2023-06-13
More accessible download links

### LINKED ISSUE #633: Web accessibility: Ensure file download links have alt text
The download links to files do not have alternative text used by screen readers, this is an accessibility requirement.

Screenshot 2023-06-12 at 3 31 10 PM

edsu:
Adds a title to the download links which will make them noticeable in a screen reader.

Here's what the screen looks like, with the downloadable file:

Screenshot 2023-06-13 at 5 14 01 PM

Here's what it sounded like in Voice Over on a Mac before the title:

https://github.com/sul-dlss/libsys-airflow/assets/33829/c0dfa2da-ab4c-43a8-87d9-7b9637de0ac7

And here's what it sounds like afterwards--notice that when reading the link it now says "Download original file":

https://github.com/sul-dlss/libsys-airflow/assets/33829/748fada1-b6f2-46d0-80e7-c9f7c14c3718

Unfortunately there is no alt-text for the bootstrap3 glyphicons which use `<span>` elements that do not have `alt` parameters.

Resolves #633

### DIFF
--- a/libsys_airflow/plugins/vendor_app/templates/vendors/_macros.html
+++ b/libsys_airflow/plugins/vendor_app/templates/vendors/_macros.html
@@ -1,9 +1,15 @@
 {% macro fileDownload(file) -%}
   {% if file.status != 'purged' %}
     (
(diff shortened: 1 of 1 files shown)


aaron-collier (approved):

aaron-collier:
I really appreciate the inclusion of the screen reader recording - that really shows how hard it is without this text.

PR #651, opened 2023-06-13
Improve accessibility of vendor app (sans file page)

### LINKED ISSUE #634: Web accessibility: Dashboard
There are a few basic accessibility issues to address on the dashboard page:

- [x] set `role='heading'` here https://github.com/sul-dlss/libsys-airflow/blob/main/libsys_airflow/plugins/vendor_app/templates/vendors/dashboard.html#L6
- [x] Structure headings. The page heading is an `h1` element, then "In Progress" and "Errors" are `h3`, however they should follow the structure and be `h2` elements, or the page heading should be shifted to an `h2`.
- [x] All of the text areas (titles, tab headings, etc) need to be included in an aria landmark: https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/region.html
- [x] Tab headings need appropriate role organization (https://www.w3.org/WAI/standards-guidelines/act/rules/bc4a75/proposed/) - this seems tenuous, the SiteImprove message is "Make sure required-owned elements are nested correctly under the context-giving element.
- [x] Ensure color contrasts are at least 7 (per SiteImprove)

Note: It may be easiest to understand the above points if you install the SiteImprove extension (https://www.siteimprove.com/integrations/browser-extensions) in order to see the specifics (especially on the color contrast)

mjgiarlo:
Fixes #634

This commit resolves the accessibility dings within every page in the vendor app other than the file page which is handled by #652 . All remaining dings are either WCAG AAA or in the Airflow UI which is outside our direct control.

### DIFF
--- a/libsys_airflow/plugins/vendor_app/templates/vendors/dashboard.html
+++ b/libsys_airflow/plugins/vendor_app/templates/vendors/dashboard.html
@@ -2,76 +2,96 @@
 {% import "vendors/_macros.html" as _macros -%}
 
 {% block content %}
-<div class="page-header">
+<header class="page-header">
   <h1>Vendor Management</h1>
-</div>
+</header>
 
-<ul class="nav nav-tabs">
-  <li role="presentation" class="active"><a href="{{ url_for('VendorManagementView.dashboard') }}">Dashboard</a></li>
-  <li role="presentation"><a href="{{ url_for('VendorManagementView.vendors') }}">Vendors</a></li>
-</ul>
+<nav>
+  <ul class="nav nav-tabs">
+    <li class="active"><a href="{{ url_for('VendorManagementView.dashboard') }}">Dashboard</a></li>
+    <li><a href="{{ url_for('VendorManagementView.vendors') }}" style="color: #555;">Vendors</a></li>
+  </ul>
+</nav>
 
-<div class="panel panel-success">
-  <div class="panel-heading"><h3>In Progress</h3></div>
+<main>
+  <div class="panel panel-success">
+    <div class="panel-heading"><h2 style="color: #0b5e49;">In Progress</h2></div>
 
-  <table class="table table-striped" id="inProgressTable">
-    <thead>
-      <th>ID</th>
-      <th>Vendor</th>
-      <th>Vendor Interface</th>
-      <th>Filename</th>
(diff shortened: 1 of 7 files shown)


libsys_airflow/plugins/vendor_app/templates/vendors/dashboard.html
```diff
@@ -2,76 +2,98 @@
 {% import "vendors/_macros.html" as _macros -%}
 
 {% block content %}
-<div class="page-header">
-  <h1>Vendor Management</h1>
-</div>
+<header>
```
jermnelson, line 5: Since the `<header>` is a block element, could you remove the `<div>` and just add the class *page-header* to the `<header>` element?

mjgiarlo: Good call!

libsys_airflow/plugins/vendor_app/templates/vendors/dashboard.html
```diff
@@ -2,76 +2,98 @@
...
+  <ul class="nav nav-tabs">
+    <li class="active"><a href="{{ url_for('VendorManagementView.dashboard') }}">Dashboard</a></li>
+    <li><a href="{{ url_for('VendorManagementView.vendors') }}" style="color: #555;">Vendors</a></li>
+  </ul>
+</nav>
 
-<div class="panel panel-success">
-  <div class="panel-heading"><h3>In Progress</h3></div>
+<main>
+  <div class="panel panel-success">
```
jermnelson, line 19: I guess a more general question for the team is we add a  `<main>` element to address *aria landmarks* or would using the `role="main"` for the `<div>` work better?

mjgiarlo: @jermnelson AFAICT, they are equivalent? In which case, I might like the more descriptive and more terse `<main>`. Open to other ideas, naturally.

jermnelson: Yes, understood they are equivalent. Was debating myself on what approach to take in my current branch :-) I'll go ahead and use the `<main>` element to make it consistent with this PR.

justinlittman (approved):

PR #656, opened 2023-06-14
Change logic that determines if email is enabled

### LINKED ISSUE #645: Don't enable email with local dev setup
Update the check for whether email should be considered enabled by changing the logic in https://github.com/sul-dlss/libsys-airflow/blob/8f8841aee035a17b75722db598925e8e08a3b4a5/libsys_airflow/plugins/vendor/emails.py#L22

In a local instance using `docker-compose.yaml`, `smtp_host` is `localhost` rather than None. Instead, check for whether this is localhost and don't enable in that case. 

(See https://stanfordlib.slack.com/archives/C04EZLBMLQN/p1686670337193479 for background)

mjgiarlo:
Fixes #645

This commit fixes a bug manifesting in local development where the app tries to send email even though it shouldn't. Why? Because the prior implementation was based on whether the `smtp.smtp_host` Airflow configuration option was set, and that option defaults to being set to `localhost`. Instead, determine that email should not be set if the option is set to its default value.

### DIFF
--- a/libsys_airflow/plugins/vendor/emails.py
+++ b/libsys_airflow/plugins/vendor/emails.py
@@ -19,12 +19,17 @@ from libsys_airflow.plugins.vendor.edi import invoice_count
 logger = logging.getLogger(__name__)
 
 
-def _email_enabled() -> bool:
-    return conf.has_option("smtp", "smtp_host")
+# NOTE: The default value of the `smtp.smtp_host` config value is `localhost`
+#       and we consider that an indicator that the app is running in a
+#       development container where email is disabled. Our deployed
+#       environments use the docker-compose-prod.yaml configurations, which
+#       override this value and effectively communicate that email is enabled.
+def _email_disabled() -> bool:
+    return conf.get("smtp", "smtp_host") == "localhost"
 
 
 def email_args() -> dict:
-    if not _email_enabled():
+    if _email_disabled():
         return {}
 
     return {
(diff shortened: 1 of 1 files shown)


libsys_airflow/plugins/vendor/emails.py
```diff
@@ -56,17 +61,16 @@ def files_fetched_email_task(
...
-        send_email(
-            os.getenv('VENDOR_LOADS_TO_EMAIL'),
-            f"{vendor_interface_name} ({vendor_code}) - {vendor_interface_uuid} - Daily Fetch Report ({date.today().isoformat()})",
-            _files_fetched_html_content(
-                vendor_interface_name,
-                vendor_code,
-                vendor_interface_uuid,
-                downloaded_files,
-            ),
-        )
```
mjgiarlo, line 69: No need to check in this function, since we check in the task above. This brings parity to the functions in this file, the rest of which only check for disabled email in the task function not its "do the work" function.

libsys_airflow/plugins/vendor/emails.py
```diff
@@ -19,12 +19,17 @@
...
 
 
-def _email_enabled() -> bool:
-    return conf.has_option("smtp", "smtp_host")
+# NOTE: The default value of the `smtp.smtp_host` config value is `localhost`
+#       and we consider that an indicator that the app is running in a
+#       development container where email is disabled. Our deployed
+#       environments use the docker-compose-prod.yaml configurations, which
+#       override this value and effectively communicate that email is enabled.
+def _email_disabled() -> bool:
```
mjgiarlo, line 27: Instead of a `!=` check here and a bunch of `not _email_enabled()` below, use more positive logic.

aaron-collier (approved):

PR #665, opened 2023-06-14
Add 590a field to MARCit records if missing expected 035a

### LINKED ISSUE #463: Marcit specific: Set 590 brief record note
For MARCit, if a record does NOT have at least 1 035 with the value "OCoLC" then add a 590 subfield a with "MARCit brief record".

This is supported in the processing config for add files, see: https://github.com/sul-dlss/libsys-airflow/blob/main/libsys_airflow/plugins/vendor/marc.py#L85

As MARCit is the only vendor with this known use case, we have not exposed this option in the vendor management interface yet, so set this value in the processing config as shown in the above example before the `marc_processing` task of the default_processign_dag programmatically when the vendor is MARCit

mjgiarlo:
Fixes #463

### DIFF
--- a/libsys_airflow/dags/vendor/default_data_processor.py
+++ b/libsys_airflow/dags/vendor/default_data_processor.py
@@ -95,16 +95,29 @@ with DAG(
                 "920",
                 "986",
             ]  # Casts [] to defaults.
+            add_fields = []
             package_name = processing_options.get("package_name")
             if package_name:
-                params["add_fields"] = [
+                add_fields.append(
                     {
                         "tag": "910",
                         "subfields": [{"code": "a", "value": package_name}],
                     }
-                ]
-            else:
-                params["add_fields"] = None
+                )
+            if vendor_interface.vendor.vendor_code_from_folio == "sfx":
+                add_fields.append(
+                    {
+                        "tag": "590",
+                        "subfields": [{"code": "a", "value": "MARCit brief record"}],
+                        "unless": {
+                            "tag": "035",
+                            "subfields": [{"code": "a", "value": "OCoLC"}],
+                        },
+                    }
+                )
+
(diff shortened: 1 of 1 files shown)


libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -95,16 +95,29 @@ def setup():
...
+                add_fields.append(
                     {
                         "tag": "910",
                         "subfields": [{"code": "a", "value": package_name}],
                     }
-                ]
-            else:
-                params["add_fields"] = None
+                )
+            if vendor_interface.vendor.vendor_code_from_folio == "sfx":
```
mjgiarlo, line 107: Is this an OK way to determine the vendor is MARCit? Not sure whether vendor code is better than organization UUID but at least the code is already accessible here in `setup()`? 🤷🏻‍♂️

libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -95,16 +95,29 @@ def setup():
...
+                        "tag": "590",
+                        "subfields": [{"code": "a", "value": "MARCit brief record"}],
+                        "unless": {
+                            "tag": "035",
+                            "subfields": [{"code": "a", "value": "OCoLC"}],
+                        },
+                    }
+                )
+
+            params["add_fields"] = add_fields if len(add_fields) > 0 else None
```
mjgiarlo, line 119: This allows for the possibility that the vendor is MARCit *and* a user has specified a package name, otherwise returning `None`.

libsys_airflow/dags/vendor/default_data_processor.py
aaron-collier: Yes, I think code is fine as it's the only other unique field.

aaron-collier (approved):

PR #668, opened 2023-06-15
Adds environment to email subjects.

### LINKED ISSUE #462: Combine MARC processing tasks into single task utilizing processing_options configuration
We currently have multiple tasks for removing marc fields, moving marc fields, and adding marc fields. The configuration for these tasks is in the JSON configuration field, `processing_options`. 

To improving maintainability, combine all existing MARC processing tasks into a single `marc processing` airflow task. Maintaining the order that these happen in is important however so that the task first removes the configured marc fields, then moves marc fields, then adds marc fields.

justinlittman:
closes #462

### DIFF
--- a/libsys_airflow/dags/vendor/default_data_processor.py
+++ b/libsys_airflow/dags/vendor/default_data_processor.py
@@ -76,6 +76,7 @@ with DAG(
         params["download_path"] = download_path(
             params["vendor_uuid"], params["vendor_interface_uuid"]
         )
+        params["environment"] = os.getenv('HONEYBADGER_ENVIRONMENT', 'development')
 
         pg_hook = PostgresHook("vendor_loads")
         with Session(pg_hook.get_sqlalchemy_engine()) as session:
@@ -165,6 +166,7 @@ with DAG(
         processed_params["records_count"],
         job_summary["srs_stats"],
         job_summary["instance_stats"],
+        params["environment"],
     )
 
     file_not_loaded_email = file_not_loaded_email_task(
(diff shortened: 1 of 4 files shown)


libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -76,6 +76,7 @@ def setup():
         params["download_path"] = download_path(
             params["vendor_uuid"], params["vendor_interface_uuid"]
         )
+        params["environment"] = os.getenv('HONEYBADGER_ENVIRONMENT', 'development')
```
mjgiarlo, line 79: Should this be `HONEYBADGER_ENV`?

mjgiarlo, line 79: Also, can you say more about why we'd pass this value into the emails module instead of using `os.getenv` within the module?

justinlittman: The HB python library uses `HONEYBADGER_ENVIRONMENT`.

justinlittman: Trying to not tie infrastructure/environment specific details into the task code, but keep isolated in the setup.

mjgiarlo: Aha. I now see the same verbiage in the docker-compose config. Good to know, thanks!

mjgiarlo: OK, sounds good

mjgiarlo (approved):

PR #669, opened 2023-06-15
Change add_fields value matching strategy

mjgiarlo:
This commit fixes a bug with our strategy for matching values in `unless` fields within the `add_fields` data structure. We previously understood this to require an exact match, but at least for marcit 590 generation we need a substring match.

### DIFF
--- a/libsys_airflow/dags/vendor/default_data_processor.py
+++ b/libsys_airflow/dags/vendor/default_data_processor.py
@@ -108,7 +108,7 @@ with DAG(
                 add_fields.append(
                     {
                         "tag": "590",
-                        "subfields": [{"code": "a", "value": "MARCit brief record"}],
+                        "subfields": [{"code": "a", "value": "MARCit brief record."}],
                         "unless": {
                             "tag": "035",
                             "subfields": [{"code": "a", "value": "OCoLC"}],
@@ -116,7 +116,7 @@ with DAG(
                     }
                 )
 
-            params["add_fields"] = add_fields if len(add_fields) > 0 else None
+            params["add_fields"] = add_fields or None
 
             # Not yet supported in UI.
             params["archive_regex"] = processing_options.get("archive_regex")
--- a/libsys_airflow/plugins/vendor/marc.py
+++ b/libsys_airflow/plugins/vendor/marc.py
@@ -82,7 +82,7 @@ def process_marc_task(
     ]
     add_fields example: [
             { tag: "910", indicator1: '2', subfields: [{code: "a", value: "MARCit"}] },
-            { tag: "590", subfields: [{code: "a", value: "MARCit brief record"}], unless: { tag: "035", subfields: [{code: "a", value: "OCoLC"}]} },
+            { tag: "590", subfields: [{code: "a", value: "MARCit brief record."}], unless: { tag: "035", subfields: [{code: "a", value: "OCoLC"}]} },
         ]
     """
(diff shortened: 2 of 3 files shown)


libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -108,15 +108,15 @@ def setup():
                 add_fields.append(
                     {
                         "tag": "590",
-                        "subfields": [{"code": "a", "value": "MARCit brief record"}],
+                        "subfields": [{"code": "a", "value": "MARCit brief record."}],
```
mjgiarlo, line 111: Alissa pointed out this value needs a period at the end.

libsys_airflow/dags/vendor/default_data_processor.py
```diff
@@ -108,15 +108,15 @@ def setup():
...
                         "unless": {
                             "tag": "035",
                             "subfields": [{"code": "a", "value": "OCoLC"}],
                         },
                     }
                 )
 
-            params["add_fields"] = add_fields if len(add_fields) > 0 else None
+            params["add_fields"] = add_fields or None
```
mjgiarlo, line 119: Trivial refactor for simplicity. Functionally equivalent to what was there before.

libsys_airflow/plugins/vendor/marc.py
```diff
@@ -304,7 +304,7 @@ def _field_match(field: MarcField, check_field: pymarc.Field):
         return False
     for subfield in field.subfields:
         check_subfield_value = ''.join(check_field.get_subfields(subfield.code))
-        if not check_subfield_value or check_subfield_value != subfield.value:
+        if not check_subfield_value or (subfield.value not in check_subfield_value):
```
mjgiarlo, line 307: Check for substring match, not exact match, as specified in marcit 590 population feature request.

tests/vendor/test_marc.py
```diff
@@ -148,14 +148,14 @@ def test_add_fields_with_unless(tmp_path, marcit_path):
...
-            if field590:
-                assert field590["a"] == "MARCit brief record"
+            field590s = record.get_fields("590")
+            assert field035 or field590s
+            if field035 and "OCoLC" in field035["a"]:
+                assert not field590s
+            if not field035 or "OCoLC" not in field035["a"]:
+                assert field590s
+            if field590s:
+                assert field590s[0]["a"] == "MARCit brief record."
```
mjgiarlo, line 158: Previously, the code above was adding 590s to records with 035a's like `(OCoLC)foobarwhatever` and it shouldn't have been. Change the tests so that they fail without the [change above](https://github.com/sul-dlss/libsys-airflow/pull/669/files#diff-bb6330099ea169eaf06faf4c534d3a7bed57c432736315fd0c90b84230694b8eR307) and now pass.

tests/vendor/test_marc.py
```diff
@@ -148,14 +148,14 @@ def test_add_fields_with_unless(tmp_path, marcit_path):
         marc_reader = pymarc.MARCReader(fo)
         for record in marc_reader:
             field035 = record["035"]
-            field590 = record["590"]
```
mjgiarlo, line 151: For the two records in the file we're testing against, one should get the 590 and the other shouldn't. Since one won't have the 590, we can't use the dict-like accessor here else a `KeyError` is raised. Hence the switch to `get_fields()` below.

justinlittman (approved):

PR #671, opened 2023-06-15
Adds rough outline of a payments DAG.

shelleydoljack:


jermnelson (approved):
Approving as a base-line with follow-up tickets for such things a testing and parallelization of invoice processing, and other additions. Could also look at refactoring this DAG to use Taskflow patterns as well.

PR #676, opened 2023-06-16
Use TaskFlow to Parallelize Invoice Processing

### LINKED ISSUE #674: create parallel tasks for gathering data for invoices awaiting payment
We have to hit multiple endpoints to gather all the data needed to create the feeder file to send to Accounts Payable ([data mapping](https://docs.google.com/spreadsheets/d/15vSge029GYl3C9_g_izyZFbGRKt8a5j6BMBu8XETtUI/edit?usp=sharing)). This has to be done for each voucher returned from the query to `/voucher/vouchers`. The task `get_invoices_awaiting_payment` should be split out into multiple tasks, that will run in parallel for each voucher returned. For example, for each voucher, we need to get the invoiceId to do a GET to `/invoice/invoices{id}` to get some data there and then also GET to `/invoice/invoice-lines?query=invoiceId=={id}` to get the invoice lines, and from there, more data for more queries to organizations and order/order-lines endpoints.

jermnelson:
Fixes #674

- Refactors to use TaskFlow pattern for the DAG
- Most downstream tasks still stubs

Screenshot 2023-06-16 at 10 30 51 AM

### DIFF
--- a/libsys_airflow/dags/payments_to_orafin.py
+++ b/libsys_airflow/dags/payments_to_orafin.py
@@ -3,37 +3,20 @@ from datetime import datetime, timedelta
 
 from airflow import DAG
 
-from airflow.operators.python import PythonOperator
-from airflow.models import Variable
 from airflow.timetables.interval import CronDataIntervalTimetable
 
-from folioclient import FolioClient
-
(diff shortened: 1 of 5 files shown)


shelleydoljack (approved):
This is a great start!

PR #684, opened 2023-06-20
Optional type checking with mypy

jmartin-sul:
* adds dependencies to allow type checking if people wish
* adds a workflow to CI to run type checking (not necessarily intended to be a required check, seems worth discussing what seems most useful to people)
* fixes up our code so that it passes type checking
* updates readme with relevant info for running the type checker
* some other opportunistic fixups i noticed along the way

starting this off as a draft PR for discussion, since i'm not sure we want to do this?  mypy was useful for finding issues in dlme-airflow and ils-middleware, and i played with this just a bit during the WC, but mostly tried to stay away from it as it didn't seem like critical path work at the time.  but this did seem like a good experiment to finish off during the first day and a half of unscheduled week.

### DIFF
--- a/.github/workflows/typechecking.yml
+++ b/.github/workflows/typechecking.yml
@@ -1,0 +1,25 @@
+name: Type Checking
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.10'
+      - name: Install Airflow + dependencies
+        run: |
+          pip install -r requirements.txt
+          poetry install
+      - name: Type checking with mypy
+        run: poetry run mypy libsys_airflow tests
--- a/libsys_airflow/plugins/folio/helpers/marc.py
+++ b/libsys_airflow/plugins/folio/helpers/marc.py
@@ -64,7 +64,7 @@ def _add_srs_audit_record(record: dict, connection, record_type):
     return record_id
 
 
-def _check_add_srs_records(**kwargs):
+def _check_add_srs_records(**kwargs) -> None:
     srs_record: dict = kwargs["srs_record"]
     snapshot_id: str = kwargs["snapshot_id"]
     folio_client = kwargs["folio_client"]
@@ -177,11 +177,11 @@ def _get_snapshot_id(folio_client):
     return snapshot_id
 
 
-def _move_001_to_035(record: pymarc.Record) -> str:
+def _move_001_to_035(record: pymarc.Record) -> str | None:
     all001 = record.get_fields("001")
 
     if len(all001) < 1:
-        return
+        return None
 
     catkey = all001[0].data
 
@@ -198,7 +198,7 @@ def _move_001_to_035(record: pymarc.Record) -> str:
     return catkey
 
 
-def _srs_check_add(**kwargs):
+def _srs_check_add(**kwargs) -> int:
     """
     Runs audit/remediation for a single SRS file
     """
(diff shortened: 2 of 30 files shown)


.github/workflows/typechecking.yml
```diff
@@ -0,0 +1,27 @@
+# github action workflow to run mypy typechecker on our python code
+
```
mjgiarlo, line 2: I'm thinking we could drop this. Between the filename and the contents a few lines down, there's probably enough context here already even for a bleary-eyed FR.

libsys_airflow/plugins/folio/helpers/marc.py
```diff
@@ -177,11 +177,11 @@ def _get_snapshot_id(folio_client):
     return snapshot_id
 
 
-def _move_001_to_035(record: pymarc.Record) -> str:
+def _move_001_to_035(record: pymarc.Record) -> str | None:
```
mjgiarlo, line 180: Is this equivalent to `Optional[str]`? As a python type-checking newbie, I don't have a good sense of this.

libsys_airflow/plugins/vendor/download.py
```diff
@@ -49,13 +49,14 @@ def __init__(self, hook: SFTPHook, remote_path: str):
...
+        return list(self._file_descriptions.keys())
 
     def get_mod_time(self, filename: str) -> datetime:
         mod_time_str = self._file_descriptions[filename]["modify"]
-        return datetime.strptime(mod_time_str, "%Y%m%d%H%M%S")
+        # TODO: revisit this type ignore if SFTPHook.describe_directory gets better TypedDict hints, see https://peps.python.org/pep-0589/
+        return datetime.strptime(mod_time_str, "%Y%m%d%H%M%S")  # type: ignore
 
-    def get_size(self, filename: str) -> int:
+    def get_size(self, filename: str) -> int | str | None:
```
mjgiarlo, line 59: Same as `Optional[Union[str, int]]`?

mjgiarlo (approved):
I have some likely naïve questions about optionals but I dig it!

.github/workflows/typechecking.yml
jmartin-sul: ah, yeah, agree.  included only because i was cribbing off of other github workflow definitions that had this.

libsys_airflow/plugins/folio/helpers/marc.py
jmartin-sul: i think that's right.  no idea which is considered more pythonic.  my preference would be for whatever's idiomatic, but if both approaches are equal on that front, i like this just a _tiny_ bit more than `Optional`, because it keeps from having to import `Optional`, and is a little more terse.  that said, i can see the argument that `Optional` might do a better job of expressing intent? 🤷

libsys_airflow/plugins/vendor/download.py
jmartin-sul: i believe so

libsys_airflow/plugins/vendor/download.py
```diff
@@ -182,7 +184,7 @@ def _download_filepath(download_path: str, filename: str) -> str:
 
 def _record_vendor_file(
     filename: str,
-    filesize: int,
+    filesize: int | str | None,
```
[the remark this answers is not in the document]
jmartin-sul: same answer 🙂

.github/workflows/typechecking.yml
jmartin-sul: boilerplate removed

libsys_airflow/plugins/folio/helpers/marc.py
mjgiarlo: Oh, would that require a new import? If so, I agree fully.

jmartin-sul: yup, indeed, tested on the REPL to be sure after you'd asked

mjgiarlo: Thanks for checking. 👍🏻

mjgiarlo:
@jermnelson Do you want to take a look?

jermnelson (approved):

PR #698, opened 2023-07-06
Assigns HoldingsTypeId for TSV Holdings

jermnelson:
Fixes https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/291

Also requires PR https://github.com/sul-dlss/folio_migration/pull/98 to be merged.

### DIFF
--- a/libsys_airflow/plugins/folio/holdings.py
+++ b/libsys_airflow/plugins/folio/holdings.py
@@ -59,6 +59,39 @@ def _generate_lookups(holdings_tsv_path: pathlib.Path) -> dict:
     return all_holdings
 
 
+def _add_holdings_type_ids(**kwargs):
+    """
+    Looks-up and assigns Holdings Id based on FORMAT value in tsv
+    """
+    airflow = kwargs.get("airflow")
+    dag_run_id = kwargs.get("dag_run_id")
+    folio_client = kwargs.get("folio_client")
+    holdings_type_result = folio_client.folio_get("/holdings-types?limit=200")
+    holdings_type_lookup = {}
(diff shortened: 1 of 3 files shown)


libsys_airflow/plugins/folio/holdings.py
```diff
@@ -59,6 +59,39 @@ def _generate_lookups(holdings_tsv_path: pathlib.Path) -> dict:
...
 
 
+def _add_holdings_type_ids(**kwargs):
+    """
+    Looks-up and assigns Holdings Id based on FORMAT value in tsv
+    """
+    airflow = kwargs.get("airflow")
+    dag_run_id = kwargs.get("dag_run_id")
+    folio_client = kwargs.get("folio_client")
+    holdings_type_result = folio_client.folio_get("/holdings-types?limit=200")
```
shelleydoljack, line 69: Might want to add `source=="local"` if you only want the ones that will be in `symphony_holdings_types_map`.

shelleydoljack (approved):

libsys_airflow/plugins/folio/holdings.py
shelleydoljack: actually, maybe the source should be "migration" as we have done with the item note types. if you don't need to limit based on source, then don't bother with this, otherwise, i'm going to do source = migration for https://github.com/sul-dlss/folio-tasks/issues/228

jermnelson: Thanks @shelleydoljack, I don't think we need to limit based on source. I'll go ahead and merge now.

PR #705, opened 2023-07-11
Support for Instance Stat Codes tsv File

### LINKED ISSUE #563: Add instance stat codes to records for symphony item records that don't get folio item records
We think this stat code still needs to be created on folio-test?  @ahafele 

DATABASE is an instance level stat code

jermnelson:
Fixes #563

### DIFF
--- a/libsys_airflow/plugins/folio/instances.py
+++ b/libsys_airflow/plugins/folio/instances.py
@@ -56,6 +56,8 @@ def _generate_record_lookups(
                     continue
                 catkey = f"a{row['CKEY']}"
                 stat_codes = []
+                if row["ITEM_TYPE"] in lookup_stat_codes:
+                    stat_codes.append(lookup_stat_codes[row["ITEM_TYPE"]])
                 if row["ITEM_CAT1"] in lookup_stat_codes:
                     stat_codes.append(lookup_stat_codes[row["ITEM_CAT1"]])
                 if catkey in record_lookups:
--- a/libsys_airflow/plugins/folio/instances.py
+++ b/libsys_airflow/plugins/folio/instances.py
@@ -16,12 +16,14 @@ from libsys_airflow.plugins.folio.helpers import post_to_okapi, setup_data_loggi
 logger = logging.getLogger(__name__)
 
 
-def _generate_record_lookups(base_tsv: Path, lookup_stat_codes: dict) -> dict:
+def _generate_record_lookups(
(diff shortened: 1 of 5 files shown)


tests/test_instances.py
```diff
@@ -86,7 +89,17 @@ def test_adjust_records(mock_file_system, mock_dag_run):  # noqa
         ]:
             fo.write(f"{line}\n")
 
+    instatcode_tsv = mock_file_system[1] / "instatcode.tsv"
+
+    with instatcode_tsv.open("w+") as fo:
+        for line in [
+            "\ufeffCKEY\tITEM_TYPE\tITEM_CAT1",
```
jermnelson, line 96: Was seeing this utf-8 code in the `CKEY` column header in some of the instatcode.tsv files, fixed by explicitly the `fieldnames` in the reader so it doesn't matter if fixed in the future.

shelleydoljack: Ugh, BOM. Weird it is only in some of them and not all of them. The code that is generating these files is the same for each ckey range. I won't worry about it since you've got a fix for it. :)

libsys_airflow/plugins/folio/instances.py
```diff
@@ -41,6 +43,29 @@ def _generate_record_lookups(base_tsv: Path, lookup_stat_codes: dict) -> dict:
...
+                catkey = f"a{row['CKEY']}"
+                stat_codes = []
+                if row["ITEM_CAT1"] in lookup_stat_codes:
+                    stat_codes.append(lookup_stat_codes[row["ITEM_CAT1"]])
+                if catkey in record_lookups:
+                    record_lookups[catkey]["stat_codes"].extend(stat_codes)
+                else:
+                    record_lookups[catkey] = {
+                        "stat_codes": stat_codes,
+                        "suppress": False,  # discoverySuppress as default
```
shelleydoljack, line 66: DATABASE is an item type so need to also evaluate ITEM_TYPE column. And ckey 10361029 has DATABASE  item type and MARCIVE item cat1.

shelleydoljack (changes requested):

libsys_airflow/plugins/folio/instances.py
jermnelson: Thanks @shelleydoljack, latest commit now adds `ITEM_TYPE` if found in the lookup_stat_codes along with a test using the example you provided.

shelleydoljack (approved):

PR #706, opened 2023-07-11
Sets update_hrid_setting to False for all migration tranformers

jermnelson:
To prevent updating HRID settings when migrating from Symphony.


shelleydoljack (approved):
🤞 it works!

Issue #187, opened 2022-10-31
Handle boundwiths as post-migration dag


dlrueda:
Ticket is a placeholder until we figure out what bound with data from Symphony will look like.  

Dag will need to create the links between parent and child instances (and/or holdings and/or items) in FOLIO.

shelleydoljack:
To create a bound-with relationship between the item and holding record, you use the `POST /inventory-storage/bound-with-parts` endpoint with the holdingsRecordId of the BW-CHILD and the itemId of the BW-PARENT, example:
```
{
  "holdingsRecordId": "49f9c0db-161e-5bb9-a600-7c55fac49f3f",
  "itemId": "d6d1aa74-31c8-5c03-bda5-582b15155f5e"
}
```

shelleydoljack:
Tested trying to post to the endpoint with an item UUID that doesn't exist yet in folio and it errors out.
```
{
    "errors": [
        {
            "message": "Cannot set bound_with_part.itemid = a89be179-0ee5-4117-a51a-5e804c4243ec because it does not exist in item.id.",
            "type": "1",
            "code": "-1",
            "parameters": [
                {
                    "key": "bound_with_part.itemid",
                    "value": "a89be179-0ee5-4117-a51a-5e804c4243ec"
                }
            ]
        }
    ]
}
```
So posting to folio the bound-with relationships would likely need to be done after all the bibs and items are migrated.

ahafele:
In Nolana you can now set the Holdings Type as bound-with. This is desirable for us. Could this happen here as well or should I write a new ticket? @jermnelson thoughts?

jermnelson:
I think that is doable in this post-migration DAG

jgreben:
Needs discussion:
- deterministic holdings uuid generation based on data in tsv not incremental hrid
- deterministic items uuid based on barcode OR ckey, but not concatenation of both.
- library and location of BW child: needs adding to locations.tsv (and in folio)?

shelleydoljack:
@jgreben Alissa wants us to create a holdings record with type = bound-with. @ahafele which holdings record in folio should get type=bound-with? The holdings record of the principal (BW-PARENT) or the part (BW-CHILD)? Also, there doesn't appear to be a bound-with relationship holdings type: https://folio-dev.stanford.edu/settings/inventory/holdingsTypes

ahafele:
[...]

PR #723, opened 2023-08-02
Unique deterministic generation of bw relationship UUID

### LINKED ISSUE #722: Duplicate BW Relationships causing error
Looking at the logs for the `boundwith_relationship` DAG, seeing occasional error messages like:

```
[2023-08-01, 20:27:39 PDT] {bw.py:62} ERROR - Failed to post 876cb632-d291-5653-aa93-0f834e17862c from /opt/airflow/migration/iterations/manual__2023-08-01T01:05:57.902603+00:00/results/boundwith_parts.json error: {
  "errors" : [ {
    "message" : "id value already exists in table bound_with_part: 876cb632-d291-5653-aa93-0f834e17862c",
    "type" : "1",
    "code" : "-1",
    "parameters" : [ {
      "key" : "id",
      "value" : "876cb632-d291-5653-aa93-0f834e17862c"
    } ]
  } ]
```

Looking in the boundswith_parts.json file seeing duplicate IDs:
```
{"id": "876cb632-d291-5653-aa93-0f834e17862c", "holdingsRecordId": "69a8ae77-b9e4-5688-b253-b3a37609b5ce", "itemId": "dc3e674f-1c4f-503b-a5fe-81b3b6c462fb"}
{"id": "876cb632-d291-5653-aa93-0f834e17862c", "holdingsRecordId": "5ec2cbc9-d6c3-53df-9b3a-4064eeae4fee", "itemId": "dc3e674f-1c4f-503b-a5fe-81b3b6c462fb"}
```

This is because at https://github.com/sul-dlss/libsys-airflow/blob/f2a84ef68ac3910a7f9b1afc6825402ad26da6e5/libsys_airflow/plugins/folio/holdings.py#L729-L735, the bw_id UUID generation is not unique.

jermnelson:
Fixes #722

### DIFF
--- a/libsys_airflow/plugins/folio/holdings.py
+++ b/libsys_airflow/plugins/folio/holdings.py
@@ -726,23 +726,12 @@ def boundwith_holdings(*args, **kwargs):
                 if len(row["BARCODE"]) < 1:
                     continue
 
-                bw_id = str(
-                    FolioUUID(
-                        okapi_url,
-                        FOLIONamespaces.other,
-                        row["CATKEY"] + row["BARCODE"],
-                    )
-                )
-
                 bw_parts = {
-                    "id": bw_id,
                     "holdingsRecordId": holdings_id,
                     "itemId": str(
                         FolioUUID(okapi_url, FOLIONamespaces.items, row["BARCODE"])
                     ),
                 }
-
-                logger.info(f"Writing bound-with part id {bw_id} to file")
                 bwp.write(f"{json.dumps(bw_parts)}\n")
 
     bw_tsv_path.unlink()


libsys_airflow/plugins/folio/holdings.py
```diff
@@ -730,7 +730,7 @@ def boundwith_holdings(*args, **kwargs):
                     FolioUUID(
                         okapi_url,
                         FOLIONamespaces.other,
-                        row["CATKEY"] + row["BARCODE"],
+                        row["CATKEY"] + row["BARCODE"] + row["COPY"] + row["CALL_SEQ"],
```
shelleydoljack, line 733: I'm questioning why we are setting bw_id. It is just the unique identifier to the boundwith-parts table. Is it required when doing a POST? We wouldn't be doing a PUT to it to update holdingsId or itemId so maybe we just POST without including `id`? I _think_ the columns selected here to create the bw_id would be unique but if not we'd get these errors. BARCODE will be the parent barcode, CATKEY, CALL_SEQ, and COPY would be of the bw child so those would be unique, also even if it is one of these [bw-child multiple copies](https://docs.google.com/spreadsheets/d/1OxC2NpHkG_wg6AnJxAR2XsjyWHWUVyLNG1BCt_cRs6w/edit?usp=sharing), e.g. ckey 2280808, which has 1 call_seq and 2 copies (but is weirdly not in the tsv files. uh oh).

shelleydoljack (approved):

jgreben (approved):

libsys_airflow/plugins/folio/holdings.py
jermnelson: Good point; I thought we had a requirement to make these ids deterministic but looking at the original [ticket](https://github.com/sul-dlss/libsys-airflow/issues/187) that isn't the case. I removed the id entirely (you're right, it isn't necessary according to the [API](https://s3.amazonaws.com/foliodocs/api/mod-inventory-storage/p/bound-with-part.html#inventory_storage_bound_with_parts_post)

PR #746, opened 2023-08-22
Handles encoded MARC21 UTF-8 strings

jermnelson:
Fixes #744

PR #749, opened 2023-08-24
Use pymarc default to_unicode for reading MARC records

### LINKED ISSUE #744: UTF8 conversion strategy not working for certain files
POAuxam2308023144608 - found https://sul-libsys-airflow-stage.stanford.edu/vendor_management/interfaces/62

If you download the original file and break with marcedit you see the character encoding is correct. If you download the processed file the character encoding is garbled. 

Examples of the files loaded to folio. 
Original - https://folio-test.stanford.edu/inventory/view/48d671de-2c49-4fa4-991e-22d3b671f3e8
Processed by VMA - https://folio-test.stanford.edu/inventory/view/c5aa02a4-c85d-4de2-822d-13d6ec10d30c 

Another example ST26096 found at - https://sul-libsys-airflow-stage.stanford.edu/vendor_management/interfaces/86

**Not the case for GOBI** which looks to be marc 8 encoded from vendor 

5020230804_f.mrc  found at https://sul-libsys-airflow-stage.stanford.edu/vendor_management/interfaces/42

jermnelson:
Sorta Fixes #744, the strategy used in PR https://github.com/sul-dlss/libsys-airflow/pull/746 only works partially if the vendor file doesn't have errors parsing string values as utf-8 OR if the vendor file contains mixed utf-8 and marc8 characters that we were seeing in marcit ckj loads. At this point, the majority of vendor loads have parsing errors but are meant to be encoded as utf-8 so the current code was actually ending up with garbage values trying to encode as `ascii`. 

@ahafele  and I discussed and she will ticket maybe using a `MARC Processing Rule` for MARC handling that would allow users to specify handing at the FOLIO Interface level but would also require UI changes. I left a note in the `_marc8_to_unicode` comment for later reference.

 Fun stuff with trying to best handle GIGO from external vendor.

### DIFF
--- a/libsys_airflow/plugins/vendor/marc.py
+++ b/libsys_airflow/plugins/vendor/marc.py
@@ -112,14 +112,13 @@ def process_marc(
     logger.info(f"Processing from {marc_path}")
     records = []
     with marc_path.open("rb") as fo:
-        reader = _marc_reader(fo, to_unicode=False)
+        reader = _marc_reader(fo, to_unicode=True)
         for record in reader:
             if record is None:
                 logger.info(
                     f"Error reading MARC. Current chunk: {reader.current_chunk}. Error: {reader.current_exception}"
                 )
             else:
-                record = _marc8_to_unicode(record)
                 if remove_fields:
                     record.remove_fields(*remove_fields)
                 if change_fields:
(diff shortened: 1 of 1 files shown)


shelleydoljack (approved):
always "fun" with character encoding!

jgreben (approved):

PR #756, opened 2023-08-31
Handle when no winning policy exists for circ rules

jermnelson:
Discovered by user when running circ_rules tester

### DIFF
--- a/libsys_airflow/plugins/folio/circ_rules.py
+++ b/libsys_airflow/plugins/folio/circ_rules.py
@@ -5,6 +5,7 @@ import urllib.parse
 
 import requests
 
+from typing import Union
 from airflow.operators.python import get_current_context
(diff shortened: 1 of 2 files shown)


shelleydoljack (approved):
I'm not sure what the user should see. Should they see all the policies with "none of them won"? Is that what they see already? I guess I just haven't paid attention to how this app works.

jermnelson:
> I'm not sure what the user should see. Should they see all the policies with "none of them won"? Is that what they see already? I guess I just haven't paid attention to how this app works.

Good question; right now if none of the policies match (and there isn't a "No" option) then they would see "None". In the attached screen-shot of the single scenario test, the `request` row is an example of "none of them won":

Screenshot 2023-08-31 at 3 36 59 PM

PR #761, opened 2023-09-18
Adds SFTP Task for Orafin Feeder File and sets Pending for Vouchers

### LINKED ISSUE #678: add task to sftp feeder file to AP's server
Current process uses sftp to upload the file to AP's server. Details at https://docs.google.com/document/d/1iIfLXn4yy_KXeECsE-qiu-WTW9yoHe4XfOQUZOwvFBw/edit#bookmark=id.hvezofqmv6pb

### LINKED ISSUE #679: payments end-processing tasks
Once feeder file is sent, we need to update voucher data in Folio that we've sent it to AP and send an email [(separately ticketed)](https://github.com/sul-dlss/libsys-airflow/issues/710) that the feeder file was successfully sent to AP. To indicate that the voucher was sent to AP, we want to update the voucher field disbursementNumber to “Pending”.

jermnelson:
Fixes #678 
Fixes #679 

This PR enables an SFTP connection to Orafin but is currently failing because the creditials Keypair we are currently have for the AP system was generated with as an old RSA that is not recognized by the [paramiko](https://www.paramiko.org/) dependency in Airflow (error `paramiko.ssh_exception.SSHException: encountered RSA key, expected OPENSSH key`)

AP department said that they will generate a new key pair this week which we can use to test. If that approach doesn't work, we could change the using an Airflow Bash operator and pass in `-o KexAlgorithms=diffie-hellman-group14-sha1` flag to SFTP running on the docker container.

### DIFF
--- a/libsys_airflow/plugins/folio/invoices.py
+++ b/libsys_airflow/plugins/folio/invoices.py
@@ -1,5 +1,7 @@
 import logging
 
+import httpx
+
 from airflow.decorators import task
 from airflow.models import Variable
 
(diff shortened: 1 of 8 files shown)


libsys_airflow/plugins/folio/invoices.py
```diff
@@ -31,19 +33,54 @@ def _get_ids_from_vouchers(folio_query: str, folio_client: FolioClient) -> list:
...
+    for invoice in invoice_ids:
+        logger.info(f"Processing Invoice {invoice['id']}")
+        voucher_result = folio_client.get(
+            f"/voucher-storage/vouchers?query=(invoiceId=={invoice['id']})"
+        )
+        for voucher in voucher_result.get("vouchers", []):
+            voucher_id = voucher["id"]
+            voucher["disbursementNumber"] = "Pending"
+            put_result = httpx.put(
+                f"{folio_client.okapi_url}/voucher-storage/vouchers/{voucher_id}",
```
shelleydoljack, line 50: I suppose we must update via `voucher-storage` since `voucher` will probably not let us because acq unit is a protected field. Is that correct? I worry about the unintended consequences of this. Maybe we file a jira ticket about the need for a PATCH or acq unit to not be protected.

libsys_airflow/plugins/folio/invoices.py
```diff
@@ -31,19 +33,54 @@ def _get_ids_from_vouchers(folio_query: str, folio_client: FolioClient) -> list:
...
 @task
 def invoices_awaiting_payment_task() -> list:
     """
     Retrieves all awaiting payment invoices uuids from vouchers endpoint
     """
     folio_client = _folio_client()
-    sul_invoices = "(acqUnitIds==*\"bd6c5f05-9ab3-41f7-8361-1c1e847196d3\"*) and exportToAccounting=true and status=\"Awaiting payment\" and cql.allRecords=1 not disbursementNumber=\"\""
-    law_invoices = "(acqUnitIds==*\"556eb26f-dbea-41c1-a1de-9a88ad950d95\"*) and exportToAccounting=true and status=\"Awaiting payment\" and cql.allRecords=1 not disbursementNumber=\"\""
+    sul_invoices = "(acqUnitIds==*\"bd6c5f05-9ab3-41f7-8361-1c1e847196d3\"*) and exportToAccounting=true and status=\"Awaiting payment\" and cql.allRecords=1 not disbursementNumber=\"Pending\""
+    law_invoices = "(acqUnitIds==*\"556eb26f-dbea-41c1-a1de-9a88ad950d95\"*) and exportToAccounting=true and status=\"Awaiting payment\" and cql.allRecords=1 not disbursementNumber=\"Pending\""
```
shelleydoljack, line 74: This doesn't make sense. The updated to "Pending" occurs after we send invoices to AP so updating this query here means we won't be able to find invoices to put in the feeder file for the next run.

shelleydoljack (changes requested):

libsys_airflow/plugins/folio/invoices.py
jermnelson: Yes, I was seeing errors when trying to retrieve and then put a voucher using the BL endpoint when the voucher was missing an acqunit.  It was weird that even the GET was failing with the `/voucher/vouchers/{voucher_id}` if the voucher was missing the acqunit (I was able to replicate in the FOLIO UI for these invoices, the voucher wouldn't appear but the voucher-lines would)

libsys_airflow/plugins/folio/invoices.py
shelleydoljack: Ahh okay. Hopefully we won't end up with vouchers of status "awaiting payment" but have a disbursementNumber from the AP report we process.

shelleydoljack (approved):

Issue #680, opened 2023-06-16
need DAG for consuming report from AP


shelleydoljack:
Create a rough outline of the processing of the payment report we will get back from AP. 

Sample header of return report:
`Supplier Number,Supplier Name,Payment Number,Payment Date,Payment Amount,Invoice Number,Invoice Date,Invoice Amount,Paid Amount,PO Number`

The return report will be csv and include data that will allow us to update these voucher fields in Folio:
- Set disbursementDate to AP payment date
- Set disbursementNumber to AP check number
- Set amount to AP paid amount

And these invoice fields in Folio:
- Set status to Paid (does this also update voucher status to Paid when doing a PUT to /invoice/invoices? or do we update voucher status first and that will update invoice status to paid?)

shelleydoljack:
Not sure if you want a separate ticket or include in this one: email a report of paid invoices (one for each acq unit SUL and Law). Basically, a list of invoices that we set to Paid status (using the return report from AP).

PR #789, opened 2023-10-23
AP Report Management DAG

jermnelson:
Partial support for ticket #680, this DAG manages finding, downloading, and then removing the reports we receive back from campus AP system based on the feeder-file we upload to them. Here is the outline of the DAG:
Screenshot 2023-10-23 at 3 51 12 PM

This DAG does the following management of these AP reports primarily through the use of `ssh` and `scp` using the [BashOperator](https://airflow.apache.org/docs/apache-airflow/2.6.2/howto/operator/bash.html). 

1. The `find_files` task looks for any csv reports in the `OF1_PRD/outbound/data/` directory and returns a list of comma-separated file names
2. The `filter_files_task` looks in the local `/opt/airflow/orafin-files/reports` directory for any matching file names, and if the file name exists, the file then is excluded from processing by the follow-up DAG. The assumption is that pre-existing files have already been processed.
3. The `get_new_reports_task` just returns all of the new_reports from the previous step.
4. The `scp_reports` downloads the new reports to the `/opt/airflow/orafin-files/reports` directory
5. The `consolidate_reports_task` just combines the existing and new reports into a single list 
6. The `launch_report_processing_task` triggers the `ap_payment_report` DAG (in a forthcoming PR) that parses the payment report and updates FOLIO invoices and vouchers and generates emails for successes and failures
7. The `remove_files` task then removes the report from the AP server (confirmed with AP that this works on their end)


shelleydoljack (approved):
👍

PR #791, opened 2023-10-25
AP Report Processing

### LINKED ISSUE #680: need DAG for consuming report from AP

jermnelson:
Fixes #680 

This DAG takes an AP Report, parses, and then retrieves the invoices and vouchers to update the status to Paid for both while setting other properties for the voucher from the report.

Screenshot 2023-10-24 at 5 20 09 PM

This DAG is triggered by the `ap_report_management` DAG with the path to a locally stored AP report. 

From each row, the invoice is attempted retrieval from the `InvoiceNum` field.  If the invoice is missing or the invoice status is already *Paid* or *Cancelled* the invoice is skipped and later used in the `email_errors_task`. After all of the invoices are retrieved in the `retrieve_invoice_task`, each invoice's status is set to *Paid* and the voucher is retrieved from Folio based on the Invoice ID. Each voucher is updated based on fields in the row from the ap report and then the voucher is updated as well. All successfully updated invoices are then included in an email in the `emailed_paid_task`.

### DIFF
--- a/libsys_airflow/plugins/orafin/reports.py
+++ b/libsys_airflow/plugins/orafin/reports.py
@@ -61,6 +61,7 @@ def _retrieve_invoice(
             msg = f"Multiple invoices {','.join([invoice['id'] for invoice in invoice_result['invoices']])} found for folioInvoiceNo {folio_invoice_number}"
             logger.error(msg)
             task_instance.xcom_push(key="duplicates", value=msg)
+    return None
 
 
 def extract_rows(retrieved_csv: str) -> list:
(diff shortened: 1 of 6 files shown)


libsys_airflow/plugins/orafin/reports.py
```diff
@@ -75,3 +192,31 @@ def retrieve_reports() -> OperatorPartial:
...
+
+
+def update_voucher(voucher: dict, task_instance, folio_client: FolioClient) -> dict:
+    """
+    Updates Voucher based on row values
+    """
+    row = task_instance.xcom_pull(
+        task_ids="retrieve_invoice_task", key=voucher["invoiceId"]
+    )[0]
+    voucher["status"] = "Paid"
```
jermnelson, line 214: This may not be needed if the voucher's status is set based on the Invoice.

shelleydoljack (approved):
Looks good. One comment about updating the voucher amount. I double-checked the comment thread on the doc and realized she wants that and we don't need to report it out b/c she can see that somehow in folio.
We're going to have to refactor a bunch of stuff to split out the reports by acq unit, I think.

PR #796, opened 2023-10-30
Includes reason invoice is excluded in exclusion email

### LINKED ISSUE #786: keep track of reason for invoice exclusion
In the excluded invoice email, we group together the invoices and invoicelines by the reason they are excluded. The reasons for exclusion are:

1. invoiceline fund disbursement distribution type is amount split
2. vendor invoice number + folio invoice number is greater than 40 characters
3. the invoice date is after the date of the DAG run (sending to AP).
4. invoice has an invoice line that has sub-total $0
5. vendor accounting code does not contain "FEEDER"

We need to keep track of these somehow to pass them through to the different tasks.

shelleydoljack:
Closes #786

### DIFF
--- a/libsys_airflow/plugins/orafin/tasks.py
+++ b/libsys_airflow/plugins/orafin/tasks.py
@@ -70,11 +70,11 @@ def email_errors_task(ti=None):
 
 
 @task
-def email_excluded_task(invoices: list):
+def email_excluded_task(invoices_exlusion_reasons: list):
     folio_url = Variable.get("FOLIO_URL")
-    if len(invoices) > 0:
-        generate_excluded_email(invoices, folio_url)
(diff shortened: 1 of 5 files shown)


libsys_airflow/plugins/orafin/tasks.py
```diff
@@ -70,11 +70,11 @@ def email_errors_task(ti=None):
 
 
 @task
-def email_excluded_task(invoices: list):
+def email_excluded_task(invoices_exlusion_reasons: list):
```
jermnelson, line 73: Very minor suggestion, change variable to `invoices_exclusion_reasons` to fix misspelling

jermnelson (approved):
Approved with minor suggestion

PR #804, opened 2023-11-08
Ensure Foreign Currency Invoice Amount in Feeder-file equals Invoice Lines Totals 

jermnelson:
From testing new Foreign Currency Invoices, surfaced errors in how the Invoice amount is generated in the feeder-file due to how rounding Invoice lines totals does not necessary equal the amount for the Invoice. 

If there is a difference between the sum of the Invoice Lines amounts and the Invoice amount; uses the  sum of the Invoice Lines amounts for the feeder-file.


shelleydoljack (approved):
One suggestion to help with troubleshooting might be to create a property that stores the converted USD amounts for the foreign currency invoices. Maybe for the next PR.

PR #826, opened 2023-12-12
Boundwith DAG Improvements

### LINKED ISSUE #821: Testing feedback
- [x] Once submitted the webform returns a Run ID. This is hyper linked to the DAG info, but not all have access to this. Can we translate it to give the actual ID? Leave the link for those with access?
- [x] Errors should result in an email with error information. Right now when the DAG errors the user receives either no indication but also no email, or they receive the attached error screen with no information on what occurred. The user should be alerted to the error via email with some information about which file and/or IDs failed. 
- [x] Include submitted filename in email. 
- [ ] When you click the back button to get back to the BW webform it keeps the file selected file and now way to clear it out. There does not seem to be a submit another file button. _Note from Alissa: If you hit browse again you can overwrite the file. Might just be a staff training issue._

jermnelson:
Fixes #821

### DIFF
--- a/libsys_airflow/dags/new_boundwiths.py
+++ b/libsys_airflow/dags/new_boundwiths.py
@@ -13,6 +13,7 @@ from libsys_airflow.plugins.folio.helpers.bw import (
     create_admin_note,
     create_bw_record,
     email_bw_summary,
+    email_failure,
     post_bw_record,
 )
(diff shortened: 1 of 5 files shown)


shelleydoljack (approved):
Seems okay. 👍

PR #866, opened 2024-02-27
Select Ids using SQLExecuteQueryOperator connector

jgreben:
Refactors DAGS to use plugin function that apply sql from files, select ids based on selected sql, and save the results to vendor instance id files.

### DIFF
--- a/libsys_airflow/plugins/data_exports/instance_ids.py
+++ b/libsys_airflow/plugins/data_exports/instance_ids.py
@@ -1,0 +1,64 @@
+import csv
+import logging
+from datetime import datetime, timedelta
+from pathlib import Path
+
+from airflow.operators.python import get_current_context
+from airflow.providers.common.sql.operators.sql import SQLExecuteQueryOperator
+
+logger = logging.getLogger(__name__)
+
+
+def fetch_record_ids(**kwargs):
+    context = get_current_context()
+    params = context.get("params")
+    airflow = kwargs.get("airflow", "/opt/airflow/libsys_airflow")
+    sql_list = sql_files(params=params, airflow=airflow)
+    results = []
+
+    for idx, sqlfile in enumerate(sql_list):
+        task_id = f"postgres_query_{idx}"
+        with open(sqlfile) as sqf:
+            query = sqf.read()
+
+        results.extend(
+            SQLExecuteQueryOperator(
+                task_id=task_id,
+                conn_id="postgres_folio",
+                database=kwargs.get("database", "okapi"),
+                sql=query,
+                parameters={
+                    "from_date": params.get(
+                        "from_date", datetime.now().strftime('%Y-%m-%d')
+                    ),
+                    "to_date": (datetime.now() + timedelta(1)).strftime('%Y-%m-%d'),
+                },
+            ).execute(context)
+        )
+
+    return results
+
+
+def sql_files(**kwargs):
+    sql_path = Path(kwargs.get("airflow")) / "plugins/data_exports/sql"
+
+    return list(sql_path.glob("*.sql"))
+
+
+def save_ids_to_fs(**kwargs):
+    today = datetime.now().strftime('%Y%m%d%H')
+    airflow = kwargs.get("airflow", "/opt/airflow")
+    task_instance = kwargs["task_instance"]
+    vendor = kwargs["vendor"]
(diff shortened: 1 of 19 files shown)


libsys_airflow/plugins/data_exports/instance_ids.py
```diff
@@ -0,0 +1,48 @@
...
+        postgres_conn_id=kwargs.get("connection", "postgres_folio"),
+        database=kwargs.get("database", "okapi")
+    )
+    connection = pg_hook.get_conn()
+    cursor = connection.cursor()
+    cursor.execute(statements)
+    cursor.fetchall()
+
+
+def save_ids_to_fs(**kwargs):
```
jermnelson, line 28: You may want to test to ensure that the csv file is being written correctly to file system. You could create a mock for the `xcom_pull` function that returns the expected result from the SQL queries.

libsys_airflow/plugins/data_exports/marc.py
```diff
@@ -128,7 +128,7 @@ def fetch_marc(**kwargs) -> str:
     with instance_file.open() as fo:
         instance_reader = csv.reader(fo)
         for row in instance_reader:
-            uuid = row[1]
+            uuid = row[0]
```
jgreben, line 131: Note that the tuples that are returned from the sql queries do not include the row number.

jgreben: This change was reverted in favor of subsequent PR.

libsys_airflow/plugins/data_exports/instance_ids.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def fetch_record_ids(**kwargs):
+    context = get_current_context()
+    params = context.get("params")
+    airflow = kwargs.get("airflow", "/opt/airflow/libsys_airflow")
+
+    pg_hook = PostgresHook(
+        postgres_conn_id=kwargs.get("connection", "postgres_folio"),
+        database=kwargs.get("database", "okapi"),
```
shelleydoljack, line 20: Do we want to set `search_path=sul_mod_inventory_storage` here, per https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/_api/airflow/providers/postgres/hooks/postgres/index.html#airflow.providers.postgres.hooks.postgres.PostgresHook ? I just wonder if somehow setting the schema would abstract out the sql a bit (maybe not necessary). 🤷

libsys_airflow/plugins/data_exports/instance_ids.py
```diff
@@ -0,0 +1,90 @@
...
+logger = logging.getLogger(__name__)
+
+
+def fetch_record_ids(**kwargs):
+    context = get_current_context()
+    params = context.get("params")
+    airflow = kwargs.get("airflow", "/opt/airflow/libsys_airflow")
+
+    pg_hook = PostgresHook(
+        postgres_conn_id=kwargs.get("connection", "postgres_folio"),
```
shelleydoljack, line 19: Is this coming from a variable we store in airflow? Or maybe from vault?

shelleydoljack (commented):
Should we be using this [SQLExecuteQueryOperator class](https://airflow.apache.org/docs/apache-airflow-providers-common-sql/stable/_api/airflow/providers/common/sql/operators/sql/index.html#airflow.providers.common.sql.operators.sql.SQLExecuteQueryOperator)? It doesn't seem like we are.

libsys_airflow/plugins/data_exports/instance_ids.py
```diff
@@ -0,0 +1,90 @@
...
+    cursor.close()
+
+    return result
+
+
+def sql_files(**kwargs):
+    params = kwargs.get("params")
+    sql_path = Path(kwargs.get("airflow")) / "plugins/data_exports/sql"
+
+    return params.get("sql_files", os.listdir(sql_path))
```
jermnelson, line 37: Since `sql_path` is already a pathlib.Path, you could get a list of sql files by doing `list(sql_path.glob("*.sql"))` instead of using the `os.listdir` method. The advantage of using the `.glob` method is that you only list those files with an `sql` file suffix.

jermnelson (approved):
Left a non-blocking suggestion.

jermnelson (approved):

shelleydoljack (approved):
👍

Issue #887, opened 2024-03-08
GOBI records have "gls" in the 035 field


shelleydoljack:
https://github.com/sul-dlss/libsys-airflow/blob/a7626af5352a926057e8faab298861d2e5860a19/libsys_airflow/plugins/data_exports/marc/exports.py#L16

Example MARC for a gob-provided record: https://folio.stanford.edu/inventory/viewsource/5e9cfb70-1977-4a20-9ebf-c995180c69ff
```
{
    "id": "ceaac8e4-817b-458e-9a14-af4b5ddb3257",
    "snapshotId": "edfa95af-c1b6-4902-a8f6-ba543e64f636",
    "matchedId": "ceaac8e4-817b-458e-9a14-af4b5ddb3257",
    "generation": 0,
    "recordType": "MARC_BIB",
    "rawRecord": {
        "id": "ceaac8e4-817b-458e-9a14-af4b5ddb3257",
[...]

PR #888, opened 2024-03-08
Refactor plugins, connects selection dags to transformers

jgreben:
@jermnelson worked with me to create `Transformer` and `Exporter` classes to allow for instantiation of the folio client and other class methods, and better mocking of folio client in the tests without having to pass the client around in messages to other functions.

### DIFF
--- a/libsys_airflow/plugins/data_exports/marc/exporter.py
+++ b/libsys_airflow/plugins/data_exports/marc/exporter.py
@@ -10,127 +10,121 @@ logger = logging.getLogger(__name__)
 
 
 class Exporter(object):
-  def __init__(self):
-    self.folio_client = folio_client()
+    def __init__(self):
+        self.folio_client = folio_client()
 
+    def check_001(self, field001s: list) -> bool:
+        reject = False
+        for field in field001s:
+            if field.value() == "gls":
+                reject = True
+        return reject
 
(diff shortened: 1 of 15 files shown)


libsys_airflow/plugins/data_exports/marc/exporter.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class Exporter(object):
+    def __init__(self):
+        self.folio_client = folio_client()
+
+    def check_001(self, field001s: list) -> bool:
```
jermnelson, line 16: If you want (but definitely not required for this PR) you could change to the method to `check_035` instead of the 001 to address ticket https://github.com/sul-dlss/libsys-airflow/issues/887. The logic would be similar to the `check_580` for the 035, subfield `a`.

jermnelson (approved):
Looks good and I like how the new Classes simplifies the task functions for the DAGs. Have a small suggestion to address ticket 887 but not a blocker.

libsys_airflow/plugins/data_exports/marc/exporter.py
jgreben: Since there's a ticket we can do it in a follow-up PR.

PR #895, opened 2024-03-13
Template and View for downloading data exports MARC files

### LINKED ISSUE #855: Create View and Template for file Download
Similar to "Folio Reports and Logs" create a template and view that will expose all of the instance id lists and processed marc files for each of the vendors.

jgreben:
- Adds template and view for downloading data exports marc files
- Adds app folder for data exports apps
- Moves vendor list to json file
- Fixes #855 
- Adds downloading of transmitted marc files.

### DIFF
--- a/libsys_airflow/plugins/data_exports/apps/data_export_download_view.py
+++ b/libsys_airflow/plugins/data_exports/apps/data_export_download_view.py
@@ -8,6 +8,12 @@ vendor_file = open(parent / "vendors.json")
 vendors = json.load(vendor_file)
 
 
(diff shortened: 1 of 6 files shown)


jermnelson (approved):
Looks good; like the refactoring to DRY up both upload and download

shelleydoljack (approved):
Just want to point out that once the transmission DAGs run for each of these, the files will be moved to the transmitted folders. It seems like maybe we should account for this somehow...

jermnelson (approved):

PR #912, opened 2024-03-22
OCLC API Wrapper for New and Updated MARC records

### LINKED ISSUE #900: OCLC Transmission DAG
Taking the MARC files generated in PR #893 (update and new marc files for SUL, Hoover, Business, and Law), use the OCLC API to either [bib-create](https://developer.api.oclc.org/wc-metadata-v2#/Manage%20Bibliographic%20Records/bib-create) for new records or [bib-replace](https://developer.api.oclc.org/wc-metadata-v2#/Manage%20Bibliographic%20Records/bib-replace) for updated records.

We could look at using the [bib-match](https://developer.api.oclc.org/wc-metadata-v2#/Manage%20Bibliographic%20Records/match-bib) API endpoint to do a best match on new records to apply the OCLC Match on the MARC record and then do the [bib-replace](https://developer.api.oclc.org/wc-metadata-v2#/Manage%20Bibliographic%20Records/bib-replace) and update the MARC record in FOLIO.

jermnelson:
Fixes #900 

TODO:
- [x]  Code coverage from new tests
- [x] Lint and type hints
- [x] Look at using [BookOps Worldcat](https://github.com/BookOps-CAT/bookops-worldcat) package for OCLC API calls
- [ ] Test on airflow-dev


libsys_airflow/plugins/data_exports/oclc_api.py
```diff
@@ -0,0 +1,152 @@
...
+        self.oclc_headers = None
+        user = kwargs["user"]
+        password = kwargs["password"]
+        self.snapshot = None
+        self.__authenticate__(user, password)
+        self.folio_client = folio_client()
+
+    def __authenticate__(self, username, passphrase) -> None:
+        try:
+            result = httpx.post(
```
shelleydoljack, line 34: We should use a client instead of the helper functions, in order to take advantage of connection pooling, I think since we could be making a large number of requests in one dag run.

jermnelson: Done, see API code and tests

libsys_airflow/plugins/data_exports/oclc_api.py
```diff
@@ -0,0 +1,219 @@
...
+        return True
+
+    def new(self, marc_files: List[str]) -> dict:
+        output: dict = {"success": [], "failures": []}
+        if len(marc_files) < 1:
+            logger.info("No new marc records")
+            return output
+        marc_records = self.__read_marc_files__(marc_files)
+
+        for record in marc_records:
```
shelleydoljack, line 164: We should move the session outside the for-loop in order to not instantiate new sessions for each in the loop.

jgreben (approved):

shelleydoljack (changes requested):
Are these real? https://github.com/sul-dlss/libsys-airflow/blob/4bdb39c732b5c8ad5f2bd337cc43648eadce9225/tests/data_exports/test_oclc_api.py#L248 I tried checking vault and did not see a key for oclc. I don't remember where these are stored.

jermnelson:
> Are these real?
> 
> https://github.com/sul-dlss/libsys-airflow/blob/4bdb39c732b5c8ad5f2bd337cc43648eadce9225/tests/data_exports/test_oclc_api.py#L248
> I tried checking vault and did not see a key for oclc. I don't remember where these are stored.

Nope, they are made up for the test.

shelleydoljack (approved):

PR #916, opened 2024-03-26
Save full dump marc to S3

jgreben:
- Skips saving instance ids to files
- pulls marc for files queried by materialized view, processes, and writes to S3
- Upgrades airflow.


libsys_airflow/plugins/data_exports/full_dump_marc.py
```diff
@@ -53,7 +50,9 @@ def fetch_number_of_records(**kwargs) -> int:
         sql=query,
     ).execute(context)
 
-    return int(result)
+    count = result[0][0]
+    logger.info(f"Record count: {count}")
+    return int(result[0][0])
```
jermnelson, line 55: Could you just return `int(count)` since you defined the variable in line 53 to be `result[0][0]`?

jermnelson (approved):
Small non-blocking suggestion about using the `count` variable

libsys_airflow/plugins/data_exports/full_dump_marc.py
jgreben: Thanks, yeah meant to do that..

PR #901, opened 2024-03-15
Adds ftp and http hooks to transmission tasks

shelleydoljack:
Closes #889

Http hook not working. Per mypy:
```
libsys_airflow/plugins/data_exports/transmission_tasks.py:43: error: Argument "data" to "run" of "HttpHook" has incompatible type
"bytes"; expected "dict[str, Any] | str | None"  [arg-type]
                hook.run(headers=connection.extra_dejson, data=marc_data)
```

PR #917, opened 2024-03-27
Patches for transmitting data tasks

shelleydoljack:
I think maybe https://github.com/sul-dlss/libsys-airflow/pull/901 got merged prematurely? No matter, here's a new PR.


jermnelson (approved):
Sorry, I was the one that merged #901 prematurely!

PR #937, opened 2024-04-19
Generate isbn list for gobi

### LINKED ISSUE #872: Add function(s) for creating Gobi lists for Data Export
Gobi requires the generation of ISBN lists specified [here](https://docs.google.com/document/d/1NAbY9L6G16DItIBjtRQKFDcL6NN5cWK8VfsFCT48vTA/edit#heading=h.ug2nnjszleyr) and does not require MARC processing or transmission.

Send a text file with 3 pipe-delimited fields on each line. The fields in order are the ISBN, "print" or "ebook", and account number "325099"
_Note: some ISBNs have two lines in the export, one each for "print" or "ebook" if the record is considered both print and ebook._

568987545678|print|325099
987654567890|ebook|325099
675456787678|print|325099
675456787678|ebook|325099
…etc

File Naming: stf.{%Y%m%d}.txt

A record is considered print if it has an item.
For those in ebook list:
856/956: NO subfield ‘x’ text include ”subscribed” AND only has holdings record(s) type=electronic
The conditions for both print and ebook could be true, in which case the record would be considered both print and ebook.

jgreben:
Creates class for saving gobi isbn lists.
Fixes #872 
Uses mock client from transformer for tests

### DIFF
--- a/libsys_airflow/plugins/data_exports/marc/gobi.py
+++ b/libsys_airflow/plugins/data_exports/marc/gobi.py
@@ -48,7 +48,7 @@ class GobiTransformer(Transformer):
                     Some ISBNs are written with embedded hyphens to ignore, or additional text after a space.
                     """
                     isbn = isbn.replace("-", "")
-                    isbn = re.sub("\s.*", "", isbn)
+                    isbn = re.sub(r"\s.*", "", isbn)
                     if not re.search(
                         r"^(?=(?:\d){9}[\dX](?:(?:\D*\d){3})?$)(?:[\dX]{10}|\d{13})$",
                         isbn,
--- a/libsys_airflow/plugins/data_exports/marc/gobi.py
+++ b/libsys_airflow/plugins/data_exports/marc/gobi.py
@@ -1,0 +1,101 @@
+import ast
+import itertools
+import logging
+import pathlib
+import pymarc
+import re
+
+from libsys_airflow.plugins.data_exports.marc.transformer import Transformer
+
+logger = logging.getLogger(__name__)
+
+
+def gobi_list_from_marc_files(marc_file_list: str):
+    gobi_lists = []
+    gobi_transformer = GobiTransformer()
+    for marc_file in ast.literal_eval(marc_file_list):
+        gobi_lists.append(gobi_transformer.generate_list(marc_file=marc_file))
+
+    return gobi_lists
+
+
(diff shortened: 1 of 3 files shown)


libsys_airflow/plugins/data_exports/marc/gobi.py
```diff
@@ -0,0 +1,84 @@
...
+            )
+
+            print_list = []
+            ebook_list = []
+
+            for i, record in enumerate(marc_records):
+                if not i % 100:
+                    logger.info(f"{i:,} records processed")
+
+                isbn = record["020"].get_subfields("a")[0]
```
jermnelson, line 38: Using the pymarc record field accessor like `record["020"]` (and the same for the fields "035", "856", and "956" below) only retrieves the first instance of the field. If there multiple instances of the field, this pattern will miss them; also if the field doesn't exist in the record, a `KeyError` exception will be raised and processing will stop. If you use the method  `record.get_fields("020")`, it will retrieve all instances of the field or an empty list if the field isn't present.

jermnelson (changes requested):
Suggestion for retrieving MARC fields to handle edge cases where multiple fields for the same tag exist or missing fields.

jgreben:
@jermnelson can you take another look at this PR. I made changes that hopefully address the problem of repeatable MARC fields.

libsys_airflow/plugins/data_exports/marc/gobi.py
```diff
@@ -0,0 +1,107 @@
...
+                    Require 10 or 13 digits, except 10 digits can have "X" as last "digit".
+                    Some ISBNs are written with embedded hyphens to ignore, or additional text after a space.
+                    """
+                    isbn = isbn.replace("-", "")
+                    isbn = re.sub(r"\s.*", "", isbn)
+                    if not re.search(
+                        r"^(?=(?:\d){9}[\dX](?:(?:\D*\d){3})?$)(?:[\dX]{10}|\d{13})$",
+                        isbn,
+                    ):
+                        break
```
jermnelson, line 56: Doing a `break` here will exit the for loop and any remaining isbns will not be checked. You probable want to use `continue` that will go to next member in the isbns list.

jermnelson (changes requested):
One suggested change so we check all of the isbns.

jermnelson (approved):

Issue #944, opened 2024-04-29
VMA - files not found for interface


ahafele:
Interface has been successfully retrieving files until beginning of April. Interface configuration has not changed, DAG says [0 files found](https://sul-libsys-airflow-prod.stanford.edu/dags/data_fetcher/grid?run_id=manual__2024-04-29T15%3A38%3A57%2B00%3A00&execution_date=2024-04-29+15%3A38%3A57%2B00%3A00&tab=logs&dag_run_id=manual__2024-04-29T15%3A38%3A57%2B00%3A00&task_id=ftp_download_task). Regex is correct. 

Interface - https://sul-libsys-airflow-prod.stanford.edu/vendor_management/interfaces/43 
Should be retrieving the following - 
-rw-r--r--    1 incabook   incabook        16334 Apr 10 09:41 ST26673.mrc
-rw-r--r--    1 incabook   incabook        16016 Apr 10 09:41 ST26674.mrc
-rw-r--r--    1 incabook   incabook        17279 Apr 10 09:41 ST26677.mrc
-rw-r--r--    1 incabook   incabook        16446 Apr 10 09:41 ST26678.mrc
-rw-r--r--    1 incabook   incabook        18469 Apr 24 15:36 ST26708.mrc
-rw-r--r--    1 incabook   incabook        18349 Apr 24 15:36 ST26709.mrc
-rw-r--r--    1 incabook   incabook        17779 Apr 24 15:36 ST26710.mrc

jermnelson:
Looking at the Interface, I think the regular expression should be adjusted to `^Stanford/ST\d+.*\.mrc$`, the "Stanford" directory is being prepended to the list of files so the regular expression is failing.

ahafele:
Is this a change on our end and therefore across all interfaces? If so, most of our regexes will now be wrong and I wonder if we can revert back to how it behaved previously?

jermnelson:
I don't think it is across all of the interfaces but it may help to check those interfaces that have a *Remote Path* value to see they are working as expected with the regex.

ahafele:
I tried updating the interface with the suggested regex and got a [download error](https://sul-libsys-airflow-prod.stanford.edu/log?execution_date=2024-04-29T19%3A15%3A58%2B00%3A00&task_id=ftp_download_task&dag_id=data_fetcher&map_index=-1).

jermnelson:
[...]

PR #948, opened 2024-04-29
Handles inclusion of Remote Path in FTP filename

jermnelson:
Fixes error in #944's [comment](https://github.com/sul-dlss/libsys-airflow/issues/944#issuecomment-2083488048).

Will want @ahafele to maybe test in either dev or stage first before merging and deploying to production just in case there are additional issues in other parts of either the VMA or data-export.

### DIFF
--- a/libsys_airflow/plugins/vendor/archive.py
+++ b/libsys_airflow/plugins/vendor/archive.py
@@ -1,5 +1,5 @@
 import logging
-import os
+import pathlib
 import shutil
 from datetime import date
 
@@ -60,7 +60,6 @@ def archive_file(
     )
     archive_filepath = _filepath(archive_path, vendor_file.vendor_filename)
     print(f"Archive path: {archive_filepath}")
-    os.makedirs(archive_path, exist_ok=True)
     shutil.copyfile(download_filepath, archive_filepath)
     vendor_file.archive_date = date.today()
     session.commit()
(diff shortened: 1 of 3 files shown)


ahafele:
I tested this for a couple of interfaces on stage. For some interfaces (like the [one](https://sul-libsys-airflow-stage.stanford.edu/vendor_management/interfaces/139) that prompted this report) the remote path name now needs to be included in the regex, for others they are working as-is. How can we instruct staff to know when to include the remote path in the regex for file fetching - for existing and future interfaces?

ahafele:
We have another example of an interface that has stopped fetching files - https://sul-libsys-airflow-prod.stanford.edu/vendor_management/interfaces/45

ahafele:
I tested the new vendor that was reported today and it appears to be the same issue. When I update the regex to include the pathname then the files are retrieved - https://sul-libsys-airflow-stage.stanford.edu/vendor_management/interfaces/145

Though the files are retrieved they are not being processed and I don't see in the DAGs any info about why (no errors etc.) You can see in the test from yesterday that the files have not been processed and have a status of _loading_
https://sul-libsys-airflow-stage.stanford.edu/vendor_management/interfaces/139

jermnelson:
Keeping this PR around until this PR #957 is tested and reviewed to preserve the comments.

jermnelson:
closing in favor of merged PR #957

PR #949, opened 2024-05-01
Fix queries for materialzed view

jgreben:
To get unique records_lb  rows I found out about the sql `select distinct on`. From https://www.postgresql.org/docs/current/sql-select.html:

`SELECT DISTINCT ON ( expression [, ...] ) keeps only the first row of each set of rows where the given expressions evaluate to equal. The DISTINCT ON expressions are interpreted using the same rules as for ORDER BY... Note that the “first row” of each set is unpredictable unless ORDER BY is used to ensure that the desired row appears first.` 

I updated the query [on the folio-k8s wiki](https://github.com/sul-dlss/folio-k8s/wiki/Custom-Database-Indexes).

### DIFF
--- a/libsys_airflow/plugins/data_exports/materialized_view.sql
+++ b/libsys_airflow/plugins/data_exports/materialized_view.sql
@@ -1,12 +1,14 @@
 create materialized view data_export_marc as
-select
-  I.id,
-  M.content
+select I.id, M.content
 from sul_mod_inventory_storage.instance I
-inner join sul_mod_source_record_storage.records_lb R
-  on  R.external_id = I.id
-  and R.state = 'ACTUAL'
-  and (I.jsonb->>'statusId')::uuid in (
+inner join(
+  select distinct on (external_id) external_id, id, generation
+  from sul_mod_source_record_storage.records_lb
+  where state = 'ACTUAL'
+  order by external_id, generation desc
+) R
+on R.external_id = I.id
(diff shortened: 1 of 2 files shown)


libsys_airflow/plugins/data_exports/materialized_view.sql
```diff
@@ -1,19 +1,21 @@
...
-  and R.state = 'ACTUAL'
-  and (I.jsonb->>'statusId')::uuid in (
+inner join(
+  select distinct on (external_id) external_id, id, generation
+  from sul_mod_source_record_storage.records_lb
+  where state = 'ACTUAL'
+  order by external_id, generation desc
+) R
+on R.external_id = I.id
+and (I.jsonb->>'statusId')::uuid in (
```
shelleydoljack, line 11: Oh Interesting (distinct on + order by). Does this do the same as `max(generation)`? or did that not work (it seems like it would work/be the same but maybe not?)?

shelleydoljack (approved):

libsys_airflow/plugins/data_exports/materialized_view.sql
jgreben: It probably would work with MAX but for some reason I couldn't figure out how to get the GROUP BY to work. I was probably just not selecting and grouping on the right things, but then I came across this function, which I guess for me was easier to visualize what it was doing.

PR #963, opened 2024-05-07
Process MARC as updates and deletes

jgreben:
- Splits marc files into updates and deletes folders.
- Display updates and deletes for Marc files in download view
- updates tags to call correct function for removing subfields
- adds dataTables JS to Marc downloads page
- Fixes bug where upload form skips first row (inferred as header)
- Patches transmission tasks and archiving functions to use updates and deletes paths


libsys_airflow/plugins/data_exports/instance_ids.py
```diff
@@ -6,49 +6,58 @@
...
-            ).execute(context)
-        )
+    results = {"updates": [], "deletes": []}  # type: dict
+
+    for type in ["updates", "deletes"]:
+        sql_list = sql_files(params=params, airflow=airflow, type=type)
+
+        for idx, sqlfile in enumerate(sql_list):
+            task_id = f"postgres_query_{idx}"
+            with open(sqlfile) as sqf:
```
jermnelson, line 20: Is there a reason to open and read this file every iteration of both the inner `for` loop and the outer `for` loop on line 15? This is opening and reading the file 2x{number of files in sql_list}. I think you could move open file outside both of the for loops so we're only opening/reading it once.

jermnelson (changes requested):
One small suggestion so we're not opening/reading files multiple times in for loops.

libsys_airflow/plugins/data_exports/instance_ids.py
jgreben: I think it needs to read and execute each sql file sequentially, and now it is doing it for everything in the `updates` sql folder and also for the `deletes` folder (even though there is only one sql file for deletes). If I am misinterpreting the need for this, maybe we can pair and you can help me figure it out.

jgreben: The file is a different file each time in the loop.

jermnelson (approved):

jgreben:
@shelleydoljack are we good to merge this?

PR #973, opened 2024-05-08
Fix bug

jgreben:
Missed a var name change from "type" to "kind".


jgreben:
Duplicated in https://github.com/sul-dlss/libsys-airflow/pull/974

PR #974, opened 2024-05-08
Fix lookups from item record to populate 999s

jgreben:
Fixes https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/637

After switching to SQL, some of the lookups were failing and preventing the 999s from being created.

### DIFF
--- a/tests/data_exports/test_oclc.py
+++ b/tests/data_exports/test_oclc.py
@@ -99,7 +99,10 @@ def mock_folio_client():
         if args[0].startswith("/material-types"):
             return {
                 "mtypes": [
-                    {"id": "d934e614-215d-4667-b231-aed97887f289", "name": "periodical"},
+                    {
+                        "id": "d934e614-215d-4667-b231-aed97887f289",
+                        "name": "periodical",
+                    },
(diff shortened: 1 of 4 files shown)


tests/data_exports/test_oclc.py
```diff
@@ -95,6 +95,14 @@ def mock_folio_get(*args, **kwargs):
                     {"id": 'c1a86906-ced0-46cb-8f5b-8cef542bdd00', "code": 'SUL'},
                 ]
             }
+        # Material types
+        if args[0].startswith("/material-types"):
```
jgreben, line 99: Seems odd to me to have to patch the OCLC test in this way.

jermnelson: I've just been in the habit of using `startswith` instead of using an exact match `==` as it is slightly more robust. More of python convention than anything else.

jermnelson (approved):

ahafele:
I think this is failing in a different way - see comment [here](https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/637#issuecomment-2120919994)

Issue #969, opened 2024-05-08
Deletes for POD


jgreben:
File with .mrc file extension. Record specified as deleted by using d in position 05 in the MARC Leader
At a minimum, each record should also contain a MARC 001 field identifying the record to delete
Deletes sent as MARC21 binary records may be included in a file with added and updated records

PR #975, opened 2024-05-09
Set leader position 5 to "d" for records in deletes list 

jgreben:
Fixes https://github.com/sul-dlss/libsys-airflow/issues/969

### DIFF
--- a/libsys_airflow/plugins/data_exports/marc/transforms.py
+++ b/libsys_airflow/plugins/data_exports/marc/transforms.py
@@ -132,7 +132,7 @@ def leader_for_deletes(marc_file: str, full_dump: bool):
     for i, record in enumerate(marc_records):
         try:
             record.leader = pymarc.leader.Leader(record.leader)
-            record.leader[5] = "d"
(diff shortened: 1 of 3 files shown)


libsys_airflow/plugins/data_exports/marc/transforms.py
```diff
@@ -109,6 +109,45 @@ def divide_into_oclc_libraries(**kwargs):
...
+        logger.info(f"Changing leader using AWS S3 with path: {marc_path}")
+
+    with marc_path.open('rb') as fo:
+        marc_records = [record for record in pymarc.MARCReader(fo)]
+
+    logger.info(f"Changing leader for {len(marc_records):,} records")
+
+    for i, record in enumerate(marc_records):
+        try:
+            record.leader = pymarc.leader.Leader(record.leader)
```
jermnelson, line 134: If/when this upstream PR [209](https://gitlab.com/pymarc/pymarc/-/merge_requests/209) is merged, should be able to remove this line (but it won't hurt anything if it still remains)

jermnelson (approved):

PR #983, opened 2024-05-14
Deduplicate uuids fetched from sql and handle absent uuids in 999

jgreben:
We discovered that some records already had 999 fields without subfield I, and the code was just looking for the first 999 field and silently skipping the block when no subfield I was present. This change checks all of the 999 fields for uuids and raises an excepted exception if none are found.

Also de-duplicate any uuids fetched via SQL for updates and deletes.

### DIFF
--- a/libsys_airflow/plugins/data_exports/instance_ids.py
+++ b/libsys_airflow/plugins/data_exports/instance_ids.py
@@ -1,5 +1,5 @@
-import csv
 import logging
+import numpy as np
 from datetime import datetime, timedelta
 from pathlib import Path
 from typing import Union
@@ -39,6 +39,8 @@ def fetch_record_ids(**kwargs) -> dict:
                     },
                 ).execute(context)
             )
+
+        results[kind] = list(np.unique(results[kind]))
 
     return results
 
@@ -83,9 +85,7 @@ def save_ids(**kwargs) -> Union[str, None]:
     data_path.parent.mkdir(parents=True, exist_ok=True)
 
     with open(data_path, 'w') as f:
-        writer = csv.writer(f, lineterminator='\n')
         for id in data:
-            if id:
-                writer.writerow(id)
+            f.write(f"{id}\n")
 
     return str(data_path)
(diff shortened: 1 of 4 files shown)


libsys_airflow/plugins/data_exports/instance_ids.py
```diff
@@ -27,18 +28,19 @@ def fetch_record_ids(**kwargs) -> dict:
...
+                database=kwargs.get("database", "okapi"),
+                sql=query,
+                parameters={
+                    "from_date": from_date,
+                    "to_date": to_date,
+                },
+            ).execute(context)
+
+            results[kind].extend(uuids)
+            results[kind] = list(np.unique(results[kind]))
```
jermnelson, line 43: Cool way to use np.unique!

libsys_airflow/plugins/data_exports/marc/transformer.py
```diff
@@ -61,13 +62,23 @@ def add_holdings_items(self, marc_file: str, full_dump: bool):
         with marc_path.open('rb') as fo:
             for i, record in enumerate(pymarc.MARCReader(fo)):
                 try:
-                    subfields_i = record["999"].get_subfields("i")
+                    fields_999 = record.get_fields("999")
+                    regex = re.compile(
```
jermnelson, line 66: I would move this compiled regex out of the for loop. You can instantiate prior to line 62. That way the regular expression will only be complied once and is usable for each iteration of the loop.

libsys_airflow/plugins/data_exports/marc/transformer.py
```diff
@@ -61,13 +62,23 @@ def add_holdings_items(self, marc_file: str, full_dump: bool):
...
+                        if not subfields_i:
+                            raise Exception("No uuid in subfields")
+
                     new_999s = self.add_holdings_items_fields(subfields_i)
                     record.add_field(*new_999s)
                     marc_records.append(record)
                     if not i % 100:
                         logger.info(f"{i:,} processed records")
-                except TypeError as e:
+                except Exception as e:
```
jermnelson, line 81: Good call, we want to generalize logging errors.

jermnelson (changes requested):
Looks good! Small request to optimize regex outside of the for loop.

shelleydoljack (approved):

jermnelson (approved):

PR #1008, opened 2024-05-21
Lookup campuses, check before adding to gobi export list

### LINKED ISSUE #997: For GOBI write ISBN list to correct directory
- Gobi list should write to the marc-files/new directory (see #996)
- Gobi list should only be for campus SUL
- - Use the `locations_by_library_lookup` method in the refactored Transformer class.

jgreben:
Fixes #997 

- Adds a lookup for campus id based on library location.
- Adds mocks to all tests that subclass Transformer to include mock `loccamps`

### DIFF
--- a/libsys_airflow/plugins/data_exports/marc/transformer.py
+++ b/libsys_airflow/plugins/data_exports/marc/transformer.py
@@ -19,7 +19,8 @@ class Transformer(object):
         self.holdings_type = self.holdings_type_lookup()
         self.locations = self.locations_lookup()
         self.materialtypes = self.materialtype_lookup()
-        self.library_lookup = self.locations_by_library_lookup()
+        self.library_lookup = self.library_by_locations_lookup()
+        self.campus_lookup = self.campus_by_locations_lookup()
 
     def call_number_lookup(self) -> dict:
         lookup = {}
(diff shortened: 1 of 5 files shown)


libsys_airflow/plugins/data_exports/marc/transformer.py
```diff
@@ -48,7 +49,17 @@ def materialtype_lookup(self) -> dict:
...
+        lookup = {}
+        campus_result = self.folio_client.folio_get("/location-units/campuses")
+        campus_lookup = {}
+        for campus in campus_result.get("loccamps"):
+            campus_lookup[campus['id']] = campus["code"]
+        for location in self.folio_client.locations:
+            lookup[location['id']] = campus_lookup.get(location['campusId'])
+        return lookup
+
+    def library_by_locations_lookup(self) -> dict:
```
shelleydoljack, line 62: I think we can refactor this to use the campus_by_locations_lookup (actually, I thought that was the plan) since this https://github.com/sul-dlss/libsys-airflow/blob/36c784fb77d7e5a5aeae87a8d0783773ed29c48a/libsys_airflow/plugins/data_exports/marc/oclc.py#L57-L85 is essentially trying to do the same thing (in fact, it seems campus = library code for the purposes of splitting up for OCLC).

shelleydoljack (approved):
Possible refactoring here or in subsequent PR.

jermnelson (approved):

PR #1009, opened 2024-05-21
Adds task to remove s3 files in full_dump taskflow

### LINKED ISSUE #959: delete files from S3 prior to select_all_records dag
Since the batch size and concurrent jobs configuration has an effect on the filenames of files written to S3, we would need to delete the files in S3 so that the transmission DAG doesn't send duplicate records that are spread across different files. Maybe we delete files in S3 at the beginning of the selection DAG to avoid this potential issue.

jgreben:
Fixes #959


shelleydoljack (approved):
Thank you!

jermnelson (approved):

PR #1014, opened 2024-05-22
Retry transmission for send_all_records DAG

### LINKED ISSUE #899: Fail dag run if any files failed transmission
Fail the dag run by raising an exception in another downstream task that takes the failed files list.
https://github.com/sul-dlss/libsys-airflow/blob/c7f21db2e62827532fa4aca9c2f41fea6f77de8f/libsys_airflow/plugins/data_exports/transmission_tasks.py#L38

shelleydoljack:
Fixes #899 

I'm not sure how we want to handle failures in the regular transmission DAGs. I did [this](https://github.com/sul-dlss/libsys-airflow/commit/0b8fc482a77ff9f1e71a28ff8127d9f49615182d) for pod but the transmission task, I feel, is sort of an airflow anti-pattern, where we are not raising an error for a failure (and thus having airflow do a retry) but instead appending the file that failed to a list to be added to the xcom. I played with the idea of raising an error if the dag id is not send_all_records so that airflow would retry the transmission. What do you think of that approach? Otherwise, on a day-to-day basis, the `send_<vendor>_records` DAGs would simply forgo archiving the file that failed and would retry sending it the next day when the DAG runs on schedule. Is that good enough?

### DIFF
--- a/libsys_airflow/dags/data_exports/full_dump_transmission.py
+++ b/libsys_airflow/dags/data_exports/full_dump_transmission.py
@@ -58,13 +58,26 @@ def send_all_records():
         files_params="upload[files][]",
     )
 
-    retry_files = retry_failed_files_task(vendor="full-dump", files=transmit_data["failures"])
+    retry_files = retry_failed_files_task(
+        vendor="full-dump", files=transmit_data["failures"]
+    )
 
-    retry_transmission = transmit_data_http_task(retry_files, files_params="upload[files][]",)
+    retry_transmission = transmit_data_http_task(
+        retry_files,
+        files_params="upload[files][]",
+    )
 
     email_failures = failed_transmission_email(retry_transmission["failures"])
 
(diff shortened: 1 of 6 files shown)


jgreben (approved):
I think the strategy of not archiving and leaving the files in the `marc-files` directory should be fine, but what do we do if they keep failing day after day?

shelleydoljack:
> I think the strategy of not archiving and leaving the files in the `marc-files` directory should be fine, but what do we do if they keep failing day after day?

Well, we'll get an email of the failures, so presumably we'd see that and figure something out, like check the vendor's site, move or delete the failed files, re-run from a specific date to create one file to cover the mutiple days' worth of failed files. If that sounds okay, I can add the failed_transmission_email task to the other send vendor records DAGs.

jermnelson (approved):
If you want to add the ` failed_transmission_email ` task to the other vendor transmission DAGs in this PR or in a follow-up PR, I'm fine with either approach.

PR #1027, opened 2024-05-24
Skip adding 999s for holdings and items that are discoverySuppress-ed

jgreben:
Fixes issue described here: https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/637#issuecomment-2129828584

### DIFF
--- a/libsys_airflow/plugins/data_exports/marc/transformer.py
+++ b/libsys_airflow/plugins/data_exports/marc/transformer.py
@@ -4,6 +4,8 @@ import pathlib
 import pymarc
 import re
 
+from pytest import skip
+
 from libsys_airflow.plugins.folio_client import folio_client
 from airflow.operators.python import get_current_context
 from airflow.providers.common.sql.operators.sql import SQLExecuteQueryOperator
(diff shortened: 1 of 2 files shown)


libsys_airflow/plugins/data_exports/marc/transformer.py
```diff
@@ -128,6 +128,10 @@ def add_holdings_items_fields(self, instance_subfields: list) -> list:
 
                 for holding_tuple in holdings_result:
                     holding = holding_tuple[0]
+                    if len(holding.get("discoverySuppress", "")) > 0:
+                        if bool(holding["discoverySuppress"]):
+                            continue
+
```
shelleydoljack, line 134: Admittedly, I am not familiar with the sql and when that data is processed. Given the selections in [sql>updates](https://github.com/sul-dlss/libsys-airflow/tree/bb32cf93781aa4b470b82d0b613a4fdfe60eaa68/libsys_airflow/plugins/data_exports/sql/updates) would there even be holdings or items with `discoverySuppress: true` at this point in the flow?

shelleydoljack (approved):

PR #1031, opened 2024-05-28
Add 999 with value of folio item barcode

### LINKED ISSUE #1021: POD - add item barcode to 999$i
For POD export we need to add one more subfield to the 999(s) we construct
 
999$i = item barcode

jgreben:
Fixes #1021

### DIFF
--- a/libsys_airflow/plugins/data_exports/marc/transformer.py
+++ b/libsys_airflow/plugins/data_exports/marc/transformer.py
@@ -187,6 +187,8 @@ class Transformer(object):
         return field_999
 
     def add_item_subfields(self, field_999: pymarc.Field, item: dict):
+        if 'barcode' in item:
+            field_999.add_subfield('i', item['barcode'])
         if 'materialTypeId' in item:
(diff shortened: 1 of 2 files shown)


libsys_airflow/plugins/data_exports/marc/transformer.py
```diff
@@ -187,6 +187,8 @@ def add_holdings_subfields(self, holdings: dict) -> pymarc.Field:
         return field_999
 
     def add_item_subfields(self, field_999: pymarc.Field, item: dict):
+        if 'barcode' in item:
```
jermnelson, line 190: If this 999 field contains the instance id in subfield i, this would overwrite that value with the barcode? Do we care?

jermnelson (approved):
Approving with one non-blocking question

shelleydoljack (approved):

libsys_airflow/plugins/data_exports/marc/transformer.py
jgreben: Does the transformer not add a new 999 field, and that is what gets the added subfield?

PR #1033, opened 2024-05-28
Query instance id in gobi transformer instead of isbn

jgreben:
The gobi transformer was incorrectly querying isbn. Now queries instance id. Also refactors the loops so that it only queries for holdings once per marc record and items once per holding.

### DIFF
--- a/libsys_airflow/dags/data_exports/gobi_selections.py
+++ b/libsys_airflow/dags/data_exports/gobi_selections.py
@@ -15,10 +15,6 @@ from libsys_airflow.plugins.data_exports.instance_ids import (
 from libsys_airflow.plugins.data_exports.marc.gobi import gobi_list_from_marc_files
 
 from libsys_airflow.plugins.data_exports.marc.exports import marc_for_instances
-from libsys_airflow.plugins.data_exports.marc.transforms import (
-    add_holdings_items_to_marc_files,
-    remove_fields_from_marc_files,
-)
 
 default_args = {
     "owner": "libsys",
--- a/libsys_airflow/dags/data_exports/gobi_selections.py
+++ b/libsys_airflow/dags/data_exports/gobi_selections.py
@@ -71,23 +71,6 @@ with DAG(
         op_kwargs={"vendor": "gobi"},
     )
 
-    transform_marc_record = PythonOperator(
-        task_id="transform_folio_marc_record",
-        python_callable=add_holdings_items_to_marc_files,
-        op_kwargs={
-            "marc_file_list": "{{ ti.xcom_pull('fetch_marc_records_from_folio') }}",
-            "full_dump": False,
-        },
-    )
-
-    transform_marc_fields = PythonOperator(
-        task_id="transform_folio_remove_marc_fields",
-        python_callable=remove_fields_from_marc_files,
-        op_kwargs={
-            "marc_file_list": "{{ ti.xcom_pull('fetch_marc_records_from_folio') }}"
-        },
-    )
-
     generate_isbn_list = PythonOperator(
         task_id="generate_isbn_lists",
         python_callable=gobi_list_from_marc_files,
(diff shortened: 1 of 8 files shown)


libsys_airflow/plugins/data_exports/marc/gobi.py
```diff
@@ -52,62 +71,47 @@ def generate_list(self, marc_file) -> pathlib.Path:
                         """
                         isbn = isbn.replace("-", "")
                         isbn = re.sub(r"\s.*", "", isbn)
-                        if not re.search(
+                        if re.search(
```
jermnelson, line 74: Maybe use a complied regex for ISBN before this `for` loop? (Realize that this isn't new code but more of a refactor to make the for loop more efficient)

jermnelson (approved):
Non-blocking suggestion to compile ISBN regex

libsys_airflow/plugins/data_exports/marc/gobi.py
jgreben: :+1 just pushed a commit.

libsys_airflow/dags/data_exports/gobi_selections.py
```diff
@@ -71,23 +67,6 @@
...
-    )
-
-    transform_marc_fields = PythonOperator(
-        task_id="transform_folio_remove_marc_fields",
-        python_callable=remove_fields_from_marc_files,
-        op_kwargs={
-            "marc_file_list": "{{ ti.xcom_pull('fetch_marc_records_from_folio') }}"
-        },
-    )
-
```
shelleydoljack, line 90: Sure, I guess we don't need these if the generate_isbn_lists can use the untransformed MARC from SRS and lookup holdings/items info in folio.

libsys_airflow/plugins/data_exports/marc/gobi.py
```diff
@@ -21,7 +21,8 @@ def gobi_list_from_marc_files(marc_file_list: str):
 
 
 class GobiTransformer(Transformer):
-    def generate_list(self, marc_file) -> pathlib.Path:
+    def generate_list(self, marc_file):
+
         # marc_path is data-export-files/gobi/marc-files/updates/YYYYMMDD.mrc
         marc_path = pathlib.Path(marc_file)
```
shelleydoljack, line 27: Is the marc file even used here anymore? Maybe, I guess if it is "untransformed" MARC, right?

shelleydoljack (approved):

libsys_airflow/plugins/data_exports/marc/gobi.py
jgreben: Yes, the gobi transformer looks at the 020, 856, and 956 of the marc record, but not the 999s.

jgreben:
@jermnelson @shelleydoljack I added a couple of commits to fix the archiving of gobi files: See: https://github.com/sul-dlss/libsys-airflow/pull/1033/commits/5772c700773a7e19057500fe494ac53d33c4b53f and https://github.com/sul-dlss/libsys-airflow/pull/1033/commits/42157708075d4f295a691b70bb33e1efff9a03ed

shelleydoljack:
> @jermnelson @shelleydoljack I added a couple of commits to fix the archiving of gobi files: See: [5772c70](https://github.com/sul-dlss/libsys-airflow/commit/5772c700773a7e19057500fe494ac53d33c4b53f) and [4215770](https://github.com/sul-dlss/libsys-airflow/commit/42157708075d4f295a691b70bb33e1efff9a03ed)

But this will transmit the `*.mrc` files and not just `*.txt` files. We need to figure out a different way then to only transmit the txt files but then be able to archive all mrc and txt files in the gobi directory.

shelleydoljack:
@jermnelson @jgreben see https://github.com/sul-dlss/libsys-airflow/pull/1038 as a solution to my comment on this PR.

PR #1043, opened 2024-05-31
Remove data export archived files

### LINKED ISSUE #860: DAG to remove old archives files
Daily or weekly, remove archived files older than 30 days.

Is there already something in libsys-airflow that does this? What about the `purge_archived_files` DAG? Can this be repurposed? There is also the remove_archived_migration_files. Should it be removed?

jgreben:
- Remove after 90 days 
- Using existing vma functions
  - rename vma dag
  - adds tags to vma functions
  - lowers via purge period to 90 days

Fixes #860


jermnelson (approved):

shelleydoljack (approved):
👍

PR #1048, opened 2024-06-03
Fix remove archived files bug

jgreben:
- Call the task for finding files to remove instead of the function directly.
- Fix pathlib path.

### DIFF
--- a/libsys_airflow/dags/data_exports/remove-archived.py
+++ b/libsys_airflow/dags/data_exports/remove-archived.py
@@ -6,7 +6,7 @@ from datetime import datetime, timedelta
 
 from airflow.decorators import task
 
-from libsys_airflow.plugins.vendor.purge import remove_files
+from libsys_airflow.plugins.vendor.purge import remove_downloads_task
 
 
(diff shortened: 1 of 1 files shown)


libsys_airflow/dags/data_exports/remove-archived.py
```diff
@@ -41,6 +42,6 @@ def gather_files_task(**kwargs) -> list[pathlib.Path]:
 
     gathered_files = gather_files_task()
 
-    remove_archived_files = remove_files(target_files=gathered_files)  # type: ignore
+    remove_archived_files = remove_downloads_task(gathered_files)  # type: ignore
```
shelleydoljack, line 45: Is this `# type: ignore` still needed?

shelleydoljack (approved):

jermnelson (approved):

libsys_airflow/dags/data_exports/remove-archived.py
jgreben: Yes :\

PR #1051, opened 2024-06-03
Fixes bug for archiving of GOBI files

### LINKED ISSUE #1047: Gobi transmission DAG not working
DAG is successful but files are not moved to state=transmitted

https://sul-libsys-airflow-dev.stanford.edu/dags/send_gobi_records/grid?run_id=manual__2024-06-03T15:46:22.307587%2B00:00&execution_date=2024-06-03+15:46:22.307587%2B00:00&tab=graph&dag_run_id=manual__2024-06-03T15:46:22.307587%2B00:00

shelleydoljack:
Fixes #1047 

I found myself jumping through hoops trying to address this issue on the `archive_transmitted_data_task` side. The prefix "stf." was part of the filename being looked for in the instanceids directory and related marc files to archive after the gobi ".txt" file is transmitted. I figured it would be easier if the "stf." was prepended when the file is uploaded to the GOBI ftp site since that is their spec for filenaming. We will have to do something similar for SHAREVDE anyways...

After re-writing the tests I then realized that maybe the vendor's filenaming convention should be in the extra field of the airflow Connection, like the remote_path is. If we want to go that route instead, I can make more changes.

### DIFF
--- a/libsys_airflow/plugins/data_exports/marc/gobi.py
+++ b/libsys_airflow/plugins/data_exports/marc/gobi.py
@@ -26,8 +26,8 @@ class GobiTransformer(Transformer):
         # marc_path is data-export-files/gobi/marc-files/updates/YYYYMMDD.mrc
         marc_path = pathlib.Path(marc_file)
         gobi_list_name = marc_path.stem
-        # gobi_path is data-export-files/gobi/marc-files/updates/stf.YYYMMDD.txt
-        gobi_path = pathlib.Path(marc_path.parent) / f"stf.{gobi_list_name}.txt"
+        # gobi_path is data-export-files/gobi/marc-files/updates/YYYMMDD.txt
+        gobi_path = pathlib.Path(marc_path.parent) / f"{gobi_list_name}.txt"
 
         with marc_path.open('rb') as fo:
             marc_records = [record for record in pymarc.MARCReader(fo)]
--- a/libsys_airflow/plugins/data_exports/transmission_tasks.py
+++ b/libsys_airflow/plugins/data_exports/transmission_tasks.py
@@ -208,10 +208,12 @@ def archive_transmitted_data_task(files):
         kind = Path(x).parent.name
         # original_transmitted_file_path = data-export-files/{vendor}/marc-files/new|updates|deletes/*.mrc|*.txt
         original_transmitted_file_path = Path(x)
+
         # archive_path = data-export-files/{vendor}/transmitted/new|updates|deletes
         archive_path = archive_dir / kind
         archive_path.mkdir(exist_ok=True)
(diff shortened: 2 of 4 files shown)


jgreben (commented):
I like the vendor filename spec idea. Can you point me to the lines where it is actually moving the .txt files to the transmitted folder? I see the comment about 
```
move marc files with same stem as transmitted filename (when transmitted file is not *.mrc)
```
but to me it looks like it is still just moving .mrc files. If it is moving .txt files to me it is very obfuscated.

shelleydoljack:
> I like the vendor filename spec idea. Can you point me to the lines where it is actually moving the .txt files to the transmitted folder? I see the comment about
> 
> ```
> move marc files with same stem as transmitted filename (when transmitted file is not *.mrc)
> ```
> 
> but to me it looks like it is still just moving .mrc files. If it is moving .txt files to me it is very obfuscated.

The archive task for GOBI gets a list of files that only includes the .txt files because that is what is sent to the transmit task: https://github.com/sul-dlss/libsys-airflow/blob/c664ace3093ee165542a603bb989931989e070bd/libsys_airflow/plugins/data_exports/transmission_tasks.py#L33-L36 The .txt files are moved here: https://github.com/sul-dlss/libsys-airflow/blob/c664ace3093ee165542a603bb989931989e070bd/libsys_airflow/plugins/data_exports/transmission_tasks.py#L228-L232

It is obfuscated by the fact that we create both marc and txt for gobi but only want to transmit txt files.

jgreben:
> The .txt files are moved here:
> https://github.com/sul-dlss/libsys-airflow/blob/c664ace3093ee165542a603bb989931989e070bd/libsys_airflow/plugins/data_exports/transmission_tasks.py#L228-L232
> 
> It is obfuscated by the fact that we create both marc and txt for gobi but only want to transmit txt files.

Thanks, can you please add blank lines on L210 and then on L211 and L216 in order to separate out the different paths and their comments? That block of code just makes it difficult to read what is happening there.

jgreben (changes requested):
Please just reformat the `archive_transmitted_data_task` function to make it more readable. Thanks.

jermnelson (approved):
Looks good. Regarding your question about adding the vendor's file-naming convention to the connection extrajson, since we already have separate DAGs for each vendor, we could add custom file-naming configuration to pass through to more generic transmission tasks.

shelleydoljack:
> Looks good. Regarding your question about adding the vendor's file-naming convention to the connection extrajson, since we already have separate DAGs for each vendor, we could add custom file-naming configuration to pass through to more generic transmission tasks.

Maybe ticket-worthy for iteration 2 (when we add sharevde).

jgreben (approved):

PR #1056, opened 2024-06-05
DRY up code base with shared plugins directory

jgreben:
DRY up code base
- move `libsys_airflow.plugins.vendor.purge` to `libsys_airflow.plugins.shared.purge`
- make a new module `libsys_airflow.plugins.shared.utils` and move the is_production function to that module.


tests/vendor/test_purge.py
```diff
@@ -12,7 +12,7 @@
 
 from libsys_airflow.plugins.vendor.models import FileStatus, VendorInterface, VendorFile
 
-from libsys_airflow.plugins.vendor.purge import (
+from libsys_airflow.plugins.shared.purge import (
```
jermnelson, line 15: Maybe move  `tests/vendor/test_purge.py` to `tests/shared/test_purge.py`?

jermnelson (approved):
This is a good idea and will make adding code that is shared across different modules easier to find and use.  One non-blocking suggestion is to move the test_purge file to `tests/shared/test_purge.py`

shelleydoljack (approved):
I wonder if utils.py should just be a top-level plugin like folio_client.py and not under a shared directory.

PR #1060, opened 2024-06-06
Trigger DAG run from data_export_upload_view

### LINKED ISSUE #1042: Upload Instance UUIDs for Data Export should Trigger Selection DAG
Currently the [data_export_upload_view](https://github.com/sul-dlss/libsys-airflow/blob/main/libsys_airflow/plugins/data_exports/apps/data_export_upload_view.py) creates the instanceIds file in the correct directory but then the various vendor selection DAGs do not have a way to query FOLIO for the MARC records or apply any transformations to those records in the MARC files. 

We should add a `TriggerDagRunOperator` for the vendor selection DAG with a new param with a file path to the instanceIds csv file. If this file path is present, each vendor selection DAG should have a branch operator that skips the `fetch_folio_record_ids` and `save_ids_to_file` task and goes directly to the `fetch_marc_records` (or equivalent task) task .

jgreben:
Fixes #1042


tests/apps/data_exports/test_upload_view.py
jermnelson, line 1: Like how this tests expectations in the UI

jermnelson (approved):

libsys_airflow/plugins/data_exports/apps/vendors.json
```diff
@@ -6,7 +6,6 @@
     "nielsen",
     "oclc",
     "pod",
-    "sharevde",
-    "west"
+    "sharevde"
```
shelleydoljack, line 9: Ha! Missed that one for WEST. Good catch.

tests/apps/data_exports/test_upload_view.py
```diff
@@ -0,0 +1,45 @@
...
+import pytest
+
+from conftest import root_directory
+from libsys_airflow.plugins.data_exports.apps.data_export_upload_view import (
+    DataExportUploadView,
+)
+
+
+@pytest.fixture
+def test_airflow_client():
```
shelleydoljack, line 13: "Test" in the fixture name is a bit confusing. Could it be changed to just `airflow_client`? Not sure if you want to use "mock" since it's not using mock. 🤷

shelleydoljack (approved):
👍 I didn't see that we have a test that the branching triggers a dag run when a file is uploaded. We test that the uploaded file is saved using the correct path. I wonder if we could add a test that the dag run is triggered too. Maybe a separate PR if the test is too involved.

jgreben:
> 👍 I didn't see that we have a test that the branching triggers a dag run when a file is uploaded. We test that the uploaded file is saved using the correct path. I wonder if we could add a test that the dag run is triggered too. Maybe a separate PR if the test is too involved.

FWIW the CircRulesTester view does not have a test for triggering the dag run. The BeautifulSoup framework does not provide a headless browser and does not have a way to take actions. It is just an html parser.

tests/apps/data_exports/test_upload_view.py
jgreben: I copied that naming convention from the other tests in the vendor_app and where I borrowed it from.

PR #1061, opened 2024-06-06
Adds "Deletes" to OCLC API Wrapper

### LINKED ISSUE #968: Deletes for OCLC
Divide records into three groups as specified on [#912](https://github.com/sul-dlss/libsys-airflow/pull/912) and submit API request to unset holdings if Record has one or none unique OCLC record id.

jermnelson:
Fixes #968 

Calls BookOps `holdings_unset` for MARC record deletes from FOLIO.

### DIFF
--- a/libsys_airflow/plugins/data_exports/oclc_api.py
+++ b/libsys_airflow/plugins/data_exports/oclc_api.py
@@ -156,7 +156,7 @@ class OCLCAPIWrapper(object):
             return output
 
         marc_records = self.__read_marc_files__(marc_files)
-        successful_files: set = set() 
+        successful_files: set = set()
         failed_files: set = set()
         with MetadataSession(authorization=self.oclc_token) as session:
             for record, file_name in marc_records:
@@ -195,6 +195,7 @@ class OCLCAPIWrapper(object):
             failures: set = kwargs["failures"]
 
             oclc_id = get_record_id(record)
+
             if len(oclc_id) != 1:
                 failures.add(file_name)
                 output['failures'].append(instance_uuid)
@@ -213,6 +214,7 @@ class OCLCAPIWrapper(object):
                 output['failures'].append(instance_uuid)
                 failures.add(file_name)
             else:
+                output['success'].append(instance_uuid)
                 successes.add(file_name)
 
         output = self.__oclc_operations__(
--- a/libsys_airflow/plugins/data_exports/oclc_api.py
+++ b/libsys_airflow/plugins/data_exports/oclc_api.py
@@ -145,6 +145,42 @@ class OCLCAPIWrapper(object):
         record.add_ordered_field(new_035)
         return self.__put_folio_record__(srs_uuid, record)
 
+    def delete(self, marc_files: List[str]) -> dict:
+        output: dict = {"success": [], "failures": []}
+        if len(marc_files) < 1:
+            logger.info("No marc records for deletes")
+            return output
+        marc_records = self.__read_marc_files__(marc_files)
+
+        with MetadataSession(authorization=self.oclc_token) as session:
+            for record in marc_records:
+                instance_uuid, srs_uuid = self.__record_uuids__(record)
+                if srs_uuid is None:
+                    continue
+                oclc_id = get_record_id(record)
+                if len(oclc_id) != 1:
+                    match len(oclc_id):
+
+                        case 0:
+                            logger.error(f"{srs_uuid} missing OCLC number")
+                            output['failures'].append(instance_uuid)
+
+                        case _:
+                            logger.error(f"Multiple OCLC ids for {srs_uuid}")
+                            output['failures'].append(instance_uuid)
+
+                    continue
+                try:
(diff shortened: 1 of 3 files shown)


libsys_airflow/plugins/data_exports/oclc_api.py
```diff
@@ -145,31 +145,35 @@ def __update_oclc_number__(self, control_number: str, srs_uuid: str) -> bool:
         record.add_ordered_field(new_035)
         return self.__put_folio_record__(srs_uuid, record)
 
-    def new(self, marc_files: List[str]) -> dict:
+    def __oclc_operations__(self, **kwargs) -> dict:
```
jermnelson, line 148: This method is part of DRYing the OCLCWrapper class, the public methods `new`, `update`, and `deletes` now define an inner function that is passed into the method.

jgreben (commented):
I know it's not part of this PR, but where do`worldcat_metadata_url` and `auth_url` get used? https://github.com/sul-dlss/libsys-airflow/blob/5892892bcd70e2188d9a882b61ec1fbc63ab0cf5/libsys_airflow/plugins/data_exports/oclc_api.py#L23-24

Same for the `__del__` method? https://github.com/sul-dlss/libsys-airflow/blob/5892892bcd70e2188d9a882b61ec1fbc63ab0cf5/libsys_airflow/plugins/data_exports/oclc_api.py#L34

libsys_airflow/plugins/data_exports/oclc_api.py
shelleydoljack: I'm not sure I understand why we are defining an inner function but if it's to get it to work within our DAGs then I guess that's okay.

libsys_airflow/plugins/data_exports/transmission_tasks.py
```diff
@@ -160,25 +160,40 @@ def transmit_data_ftp_task(conn_id, gather_files) -> dict:
...
 
         if len(records.get("update", [])) > 0:
             updated_result = oclc_api.update(records['update'])
             success[library].extend(updated_result['success'])
             failures[library].extend(updated_result['failures'])
+            archive.extend(updated_result['archive'])
 
-    return {"success": success, "failures": failures}
+    archive = list(set(archive))
+    return {"success": success, "failures": failures, "archive": archive}
```
shelleydoljack, line 196: What happens to the success and failures lists? Are these the transitory MARC files that we delete?

shelleydoljack (approved):

libsys_airflow/plugins/data_exports/transmission_tasks.py
jermnelson: We are currently not doing anything with either; they both contain Instance UUIDs that could be used to construct FOLIO URLs for email reports in follow-up PRs.

libsys_airflow/plugins/data_exports/oclc_api.py
jermnelson: In the previous `new` and `update` functions we were repeating the same code to set-up the output,  sets used to track files, and the `MetadataSession` session context. The new `delete` function would have needed to repeat the same boilerplate so creating the `__oclc_operations__` method allowed us to move that boilerplate to a single function and just pass in the functions that differed by operation.

jermnelson:
> I know it's not part of this PR, but where do`worldcat_metadata_url` and `auth_url` get used?
> 
> https://github.com/sul-dlss/libsys-airflow/blob/5892892bcd70e2188d9a882b61ec1fbc63ab0cf5/libsys_airflow/plugins/data_exports/oclc_api.py#L23-L24
> 
> Same for the `__del__` method?
> 
> https://github.com/sul-dlss/libsys-airflow/blob/5892892bcd70e2188d9a882b61ec1fbc63ab0cf5/libsys_airflow/plugins/data_exports/oclc_api.py#L34

Good catch, I just pushed up a commit that removes the 1worldcat_metadata_url` and `auth_url` that were left over when I wasn't using the BookOps package. The `__del__` method is necessary to close the `httpx.Client` that we're using for the the HTTP pooling of PUTs to FOLIO. Not explictly closing the Client was causing problems during testing.

jgreben:
> > https://github.com/sul-dlss/libsys-airflow/blob/5892892bcd70e2188d9a882b61ec1fbc63ab0cf5/libsys_airflow/plugins/data_exports/oclc_api.py#L34
> 
> Good catch, I just pushed up a commit that removes the `worldcat_metadata_url` and `auth_url` that were left over when I wasn't using the BookOps package. The `__del__` method is necessary to close the `httpx.Client` that we're using for the the HTTP pooling of PUTs to FOLIO. Not explictly closing the Client was causing problems during testing.

But I do not find a reference to the `__del__` method anywhere when I do a search ¯\_(ツ)_/¯ .

jgreben (changes requested):
Check that `self.__del__` is called where and if needed.

jgreben (approved):

PR #1066, opened 2024-06-07
Adds trigger_rule so skipped fetch ids from folio branch doesn't cause downstream skips

### LINKED ISSUE #1042: Upload Instance UUIDs for Data Export should Trigger Selection DAG

shelleydoljack:
Fixes #1042 

I'm not entirely sure why, but I didn't get this to work locally until I added the `Param(True|False, type="boolean")` to the params and conf dicts.

Screenshot 2024-06-07 at 4 52 00 PM

### DIFF
--- a/libsys_airflow/dags/data_exports/gobi_selections.py
+++ b/libsys_airflow/dags/data_exports/gobi_selections.py
@@ -50,7 +50,7 @@ with DAG(
             type="string",
             description="The latest date to select record IDs from FOLIO.",
         ),
-        "fetch_folio_record_ids": True,
+        "fetch_folio_record_ids": Param(True, type="boolean"),
         "saved_record_ids_kind": None,
     },
 ) as dag:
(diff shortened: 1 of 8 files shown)


shelleydoljack:
Although, now that I see it, I wonder how the fetch_marc_records_from_folio task completed successfully when I ran from localhost. 🤷

jermnelson (approved):

PR #1068, opened 2024-06-10
Fixes and adds tasks for OCLC Selection and Transmission DAGs

jermnelson:
Issues discovered while testing on libsys-airflow dev

### DIFF
--- a/libsys_airflow/plugins/data_exports/apps/data_export_upload_view.py
+++ b/libsys_airflow/plugins/data_exports/apps/data_export_upload_view.py
@@ -4,7 +4,6 @@ import pathlib
 import re
 
 from airflow.models import DagBag
-from airflow.models.param import Param
 from airflow.utils import timezone
 from airflow.utils.state import State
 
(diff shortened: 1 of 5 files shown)


libsys_airflow/plugins/data_exports/apps/data_export_upload_view.py
```diff
@@ -64,7 +63,7 @@ def _trigger_dag_run(self, vendor, kind):
             execution_date=execution_date,
             state=State.RUNNING,
             conf={
-                "fetch_folio_record_ids": Param(False, type="boolean"),
+                "fetch_folio_record_ids": False,
```
shelleydoljack, line 66: I think we need a test for this to check that the DAG is getting the right configuration or param. This didn't work in some DAG runs but then worked with Param but then didn't work again. 🤷 Is it the difference between a DAG conf versus param? https://github.com/sul-dlss/libsys-airflow/blob/b25279c821086ae57c7114d1c120bae09522481b/libsys_airflow/dags/data_exports/gobi_selections.py#L53

shelleydoljack (approved):
Seems okay. Just one non-blocking comment that maybe we should address in a different PR.

libsys_airflow/plugins/data_exports/apps/data_export_upload_view.py
jermnelson: I think `Param` is for configuration the DAG and not used in the `dag.create_dagrun` function.

PR #1075, opened 2024-06-13
Adds -M to OCLC prefix in 035 $a if doesn't exist

### LINKED ISSUE #1069: OCLC Export QA - Set holdings (updates)
Setting holdings in OCLC worked for first simple test! But the subsequent editing of the MARC record is not happening yet and I'm not sure if it is a bug or a missed requirement. 

When holdings are set in OCLC, e.g. Record has one unique OCLC record id, then do the following - 

> If the response code is 200, change the OCLC record id prefix in the instance record from (OCoLC) to (OCoLC-M) if it is not already that.
> If the response code is something other than 200, then appropriate action should be taken to resolve the problem, either by Systems or by Data Control, depending on the nature of the problem.

jermnelson:
Fixes #1069

### DIFF
--- a/libsys_airflow/plugins/data_exports/oclc_api.py
+++ b/libsys_airflow/plugins/data_exports/oclc_api.py
@@ -14,6 +14,8 @@ from libsys_airflow.plugins.data_exports.marc.transforms import oclc_excluded
 
 from libsys_airflow.plugins.shared.folio_client import folio_client
 
+from pymarc import Record
+
 logger = logging.getLogger(__name__)
 
 
@@ -38,7 +40,7 @@ class OCLCAPIWrapper(object):
             logger.error(msg)
             raise Exception(msg, e)
 
-    def __get_srs_record__(self, srs_uuid: str) -> Union[pymarc.Record, None]:
+    def __get_srs_record__(self, srs_uuid: str) -> Union[Record, None]:
         marc_json = self.folio_client.folio_get(f"/source-storage/records/{srs_uuid}")
         marc_json_handler = pymarc.JSONHandler()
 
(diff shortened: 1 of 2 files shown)


libsys_airflow/plugins/data_exports/oclc_api.py
```diff
@@ -121,24 +123,36 @@ def __update_035__(self, oclc_put_result: bytes, srs_uuid: str) -> bool:
...
-            subfields=[pymarc.Subfield(code='a', value=f"(OCoLC){control_number}")],
-        )
-        record.add_ordered_field(new_035)
-        return self.__put_folio_record__(srs_uuid, record)
+            for i, subfield in enumerate(field.subfields):
+                if subfield.code == "a" and control_number in subfield.value:
+                    # Control number already exists
+                    if subfield.value.startswith("(OCoLC)"):
+                        # Change prefix to include -M
+                        new_prefix = subfield.value.replace("(OCoLC)", "(OCoLC-M)")
```
shelleydoljack, line 143: Are the numbers the same here and all that needs to be done is the `-M` added? @ahafele I know Vitus said it's just the stuff in parentheses that need changing but do we know that the number is the same?

libsys_airflow/plugins/data_exports/oclc_api.py
```diff
@@ -121,24 +123,36 @@ def __update_035__(self, oclc_put_result: bytes, srs_uuid: str) -> bool:
...
+                        field.subfields.pop(i)
+                        field.add_subfield(code="a", value=new_prefix, pos=i)
+                    needs_new_035 = False
+                    break
+        if needs_new_035:
+            new_035 = pymarc.Field(
+                tag='035',
+                indicators=[' ', ' '],
+                subfields=[pymarc.Subfield(code='a', value=f"(OCoLC){control_number}")],
```
shelleydoljack, line 152: Need to confirm if `-M` should be added here. Is this part addressing the "new" records we add to OCLC?

shelleydoljack (commented):
I'm concerned about adding the `-M` if the number isn't the same. Maybe it is b/c the API endpoint we send in has the control number in it? The API is unfamiliar to me so maybe this is already addressed.

libsys_airflow/plugins/data_exports/oclc_api.py
jermnelson: @shelleydoljack from your comment I realized that the test that just checked if control_number from OCLC was present in the subfield `a` would return true even if the control_number was part of a larger number. ie 
"456719" would match "(OCoLC-M)456719" but also matches "(OCoLC)6456719", and "(OCoLC)4567199".  

I refactored to instead use a regular expression that extracts the `-M` if present along with subfield's control number.  The code then checks if the suffix is not present and the incoming control_number is equal to the existing subfield a control_number before doing the replacement. 

This will better ensure that the control_number from actually OCLC is present and the prefix should be changed.

libsys_airflow/plugins/data_exports/oclc_api.py
jermnelson: Yes, this would applied to only "new" records; @ahafele should this also be `OCoLC-M` and not just `OCoLC`?

libsys_airflow/plugins/data_exports/oclc_api.py
shelleydoljack: Oh good catch. That would have been bad.

shelleydoljack (approved):

PR #1074, opened 2024-06-13
Adjusts DAG parameters to handle None for saved_record_ids_kind

jermnelson:
Fixes #1073

Issue #1071, opened 2024-06-11
Logging for full dump appears to be broken


shelleydoljack:
There is logging to let us know which file an instance ID was written to:
https://github.com/sul-dlss/libsys-airflow/blob/b25279c821086ae57c7114d1c120bae09522481b/libsys_airflow/plugins/data_exports/marc/exporter.py#L134-L136

However, on the latest DAG run of select_all_records, the logs are missing for each mapped task in fetch_folio_records:
![Screenshot 2024-06-11 at 2 49 32 PM](https://github.com/sul-dlss/libsys-airflow/assets/6343464/b6f500d8-cb1c-49d7-ba6e-bef366a3c489)
![Screenshot 2024-06-11 at 2 49 59 PM](https://github.com/sul-dlss/libsys-airflow/assets/6343464/d0fd35af-c6a4-4001-814c-b4f9eead5c23)

https://sul-libsys-airflow-dev.stanford.edu/dags/select_all_records/grid?dag_run_id=manual__2024-06-04T12%3A35%3A33-07%3A00&tab=mapped_tasks&task_id=fetch_folio_records&map_index=0

jermnelson:
This isn't a "bug" in the code but how we deploy the application using capistrano. Each time we deploy a branch, the logs directory is re-created as an empty directory while the previous version does retain the logs.

shelleydoljack:
Oh right, of course. I guess we should be more weary of that if we want to go back and try to track where a record from the full dump might have been written.

PR #1076, opened 2024-06-13
Move Airflow Logs to Shared directories

jermnelson:
Would allow Airflow logs to persist between Capistrano deploys and would prevent issues like #1071. 

Also, when testing this PR on airflow-dev, discovered that I missed one place in the code where lists were being re-cast as a strings that is no longer needed because of PR #1074.


shelleydoljack (approved):
👍

PR #1087, opened 2024-06-19
Use Active SRS record ID for Updating SRS with Change Manager Endpoint

jermnelson:
Discovered while testing that the SRS UUID in the 999 field subfield s is not necessarily the correct UUID to use for the Okapi change-manager API. This refactoring retrieves the Active SRS UUID by using the Instance UUID to update the SRS record.


jgreben (approved):

shelleydoljack (approved):
Yes, the 999 $s will be the matched_id from records_lb and not necessarily the current source record ID. When the MARC changes and a new source record is created, the new source record ID changes, but the 999 $s should not (I didn't see that we update this so that's good).

PR #1096, opened 2024-06-25
OCLC Transmission DAG with match support

### LINKED ISSUE #1094: OCLC - add match check to new record flow
For new records (records with no OCLC number) Add a task to attempt to match first. 
`/manage/bibs/match` and if found, add the OCLC number
 if not found then POST to `/manage/bibs` POST to add the record to OCLC

Background: The /manage/bibs api documentation says it will add the record if it does not already exist, but it doesn't seem to actually check for the record first. We've sent a question to OCLC but given that Cornell also does a match search first, we should go ahead and add it to the DAG.

jermnelson:
Fixes #1094 

Refactors OCLC Transmission DAG by breaking out `deletes`, `updates`, `new` into separate tasks and adds new `match` task as a dependency for the `updates` task. 

New DAG diagram:
Screenshot 2024-06-25 at 4 29 01 PM

TODO:
- [x] Increase test coverage for:
  - [x] oclc_api
  - [x] transmission_tasks

### DIFF
--- a/libsys_airflow/plugins/data_exports/transmission_tasks.py
+++ b/libsys_airflow/plugins/data_exports/transmission_tasks.py
@@ -201,7 +201,7 @@ def new_to_oclc_task(connection_details: list, new_records: dict) -> dict:
 
 
 @task(multiple_outputs=True)
-def update_oclc_task(connection_details: list, update_records: dict) -> dict:
+def set_holdings_oclc_task(connection_details: list, update_records: dict) -> dict:
 
     connection_lookup = oclc_connections(connection_details)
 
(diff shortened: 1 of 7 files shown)


libsys_airflow/plugins/data_exports/transmission_tasks.py
```diff
@@ -68,26 +72,28 @@ def retry_failed_files_task(**kwargs) -> dict:
 @task(multiple_outputs=True)
 def gather_oclc_files_task(**kwargs) -> dict:
     """
-    Gets new and updated MARC files by library (SUL, Business, Hoover, and Law)
+    Gets deleted, new, and updated MARC files by library (SUL, Business, Hoover, and Law)
     to send to OCLC
     """
     airflow = kwargs.get("airflow", "/opt/airflow")
     libraries: dict = {
```
jgreben, line 79: Are these still dicts?

jgreben (approved):

Issue #1080, opened 2024-06-17
OCLC - reporting


ahafele:
In addition to the 2 general email reports in #1079 and #861, OCLC requires a number of specialized reports for staff processing. These reports should be split by campus code and sent to specific email addresses (see below and referenced here #1035) _and_ sul-unicorn-devs@lists.stanford.edu.
**During testing please just send to sul-unicorn-devs**

1. Record has more than one OCLC record id (and therefore not new, update (set holdings), or delete (unset holdings))

Subject/body:

"OCLC: multiple OCLC IDs"

- Date of DAG run
- Report (or link to report) with list of record HRIDs or UUIDs

2. Matches - if the response code is something other than 200 and not resolvable by airflow task.

Subject/body:

"OCLC: Match error"

- Date of DAG run
- Report (or link to report) with list of record HRIDs or UUIDs and 
- OCLC response and response message

3. Updates (Set holdings) - if the response code is something other than 200 and not resolvable by airflow task.
**NOTE: we are currently setting holdings in two tasks - match and set holdings - so will need to account for both in this report. 

Subject/body:

"OCLC: Set holdings error"

- Date of DAG run
- Report (or link to report) with list of record HRIDs or UUIDs and 
- OCLC response and response message

3b. Deletes (Unset holdings) - if the response code is something other than 200 or if the response code is 200 but any message other than "Unset Holdings Succeeded." or "WorldCat Holding already unset."

Subject/body:

"OCLC: Unset holdings error"

- Date of DAG run
- Report (or link to report) with list of record HRIDs or UUIDs and 
- OCLC response and response message

Example response 

[2024-07-26, 09:59:56 PDT] {oclc_api.py:281} INFO - Delete 1107750624, holdings_unset call response <Response [200]> body: b'{\n  "controlNumber" : "1107750624",\n  "requestedControlNumber" : "1107750624",\n  "institutionCode" : "6617",\n  "institutionSymbol" : "STF",\n  "firstTimeUse" : false,\n  "success" : false,\n  "message" : "Unset Holdings Failed. Local bibliographic data (LBD) is attached to this record. To unset the holding, delete attached LBD first and try again.",\n  "action" : "Unset Holdings"\n}'

[...]

PR #1138, opened 2024-08-12
OCLC Holdings Set and Unset Reports w/email

jermnelson:
Part of fix for ticket #1080. OCLC transmission DAG now has a reports-email Task Group that generates a holdings_set report for both new records and matched records along with a holdings_unset (delete) report. Emails are sent to each of the cohort libraries with links to the reports. 

Here is the overview of the OCLC Transmission DAG:
Screenshot 2024-08-12 at 3 28 22 PM

TODO:
- [x] Test with more UUIDs on airflow-dev
- [x] Confirm emails are being generated and sent from airflow-dev.

### DIFF
--- a/libsys_airflow/plugins/data_exports/email.py
+++ b/libsys_airflow/plugins/data_exports/email.py
@@ -1,7 +1,9 @@
 import logging
+import pathlib
 
 from jinja2 import Template
 
+from airflow.configuration import conf
 from airflow.decorators import task
 from airflow.models import Variable
 from airflow.utils.email import send_email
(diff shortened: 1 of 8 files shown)


libsys_airflow/plugins/data_exports/email.py
```diff
@@ -30,6 +32,71 @@ def _oclc_identifiers(multiple_codes: list, folio_url: str):
...
+                subject_line += " Business"
+
+            case "STF":
+                to_emails.insert(0, sul_email)
+                subject_line += " SUL"
+
+        if not is_production():
+            to_emails.pop(0)  # Should only send report to libsys devs
+
+        report_url = f"{airflow_url}data_export_oclc_reports/{library}/{report_type}/{report_path.name}"
```
shelleydoljack, line 92: So the email body contains links to a report. Do we have a retention policy setup for the reports? Or do they get cleared out as part of the DAG that cleans up old files?

libsys_airflow/plugins/data_exports/templates/data-export-oclc-reports/index.html
shelleydoljack, line 1: This looks neat. But it's probably going to get unwieldy when this runs daily in prod. Maybe we'll have to have separate pages for each library? Or maybe cleaning out old reports will make index.html not unwieldy. 🤷

shelleydoljack (approved):
Seems good so far.

libsys_airflow/plugins/data_exports/email.py
jermnelson: No retention policy has been set; I'll ask Alissa when she gets back. Good idea about adding these reports to the DAG that cleans up old files.

PR #1140, opened 2024-08-16
Error report and email for OCLC errors with new MARC records

jermnelson:
Part of #1080, creates and generates an error report for each library when trying to add a new MARC record to OCLC.

TODO:
- [ ] Test on airflow-dev
- [ ] Test coverage for new reports and email


libsys_airflow/plugins/data_exports/oclc_reports.py
```diff
@@ -117,12 +180,21 @@
...
     )
 )
 
 
+def _dag_run_url(dag_run) -> str:
+    airflow_url = conf.get('webserver', 'base_url')
+    if not airflow_url.endswith("/"):
+        airflow_url = f"{airflow_url}/"
+    params = urllib.parse.urlencode({"dag_run_id": dag_run.run_id})
+    return f"{airflow_url}dags/send_oclc_records/grid?{params}"
```
shelleydoljack, line 195: Would this be similar to https://github.com/sul-dlss/libsys-airflow/blob/075468530f4181d818067f7b53a73cbd80ffbf7e/libsys_airflow/plugins/data_exports/email.py#L222 ? Maybe we could refactor to make it accessible to the other email report? Perhaps a different PR.

shelleydoljack (approved):
Just a few non-blocking comments

jgreben (approved):

libsys_airflow/plugins/data_exports/oclc_reports.py
jermnelson: Sounds good. I can make a ticket.

shelleydoljack (approved):
👍

PR #1144, opened 2024-08-20
Adjust export sched

jgreben:
Getting pod and Gobi ready to run at 7pm/8pm and transmit 4 hours later. Archive files after midnight. Change schedule var name so that selection and transmission can be overridden separately.


shelleydoljack (approved):
Ok, but shouldn't the file libsys_airflow/dags/data_exports/default_schedules.json be updated too with `transmit_*_days|hours` keys?

jermnelson (approved):

PR #1147, opened 2024-08-21
Removes leading Zero if present in OCLC number

### LINKED ISSUE #1146: OCLC - adjust regular expression when extracting OCLC numbers
Older records can include leading zeros. For example, this record shows up in the multiple OCLC numbers report https://folio-test.stanford.edu/inventory/viewsource/fa783415-127a-47b2-b967-e4ff6806e9c9
OCLC ids are
035  ‡a (OCoLC)ocm**0**8652609
035  ‡a (OCoLC)8652609

Per Jeanette
ocm's that are shorter than 8 digits get the leading zeros padding them up to 8 digits. If the number is 9 digits it is ocn and and if the number is 10+ digits it is on

jermnelson:
Fixes #1146

### DIFF
--- a/libsys_airflow/plugins/data_exports/marc/oclc.py
+++ b/libsys_airflow/plugins/data_exports/marc/oclc.py
@@ -1,5 +1,6 @@
 import logging
 import pathlib
+import re
 
 import httpx
 import pymarc
@@ -7,6 +8,8 @@ import pymarc
 from libsys_airflow.plugins.data_exports.marc.transformer import Transformer
 
 logger = logging.getLogger(__name__)
+
+OCLC_NUMBER = re.compile(r'0*(\w*)')
 
 
 def archive_instanceid_csv(instance_id_csvs: list):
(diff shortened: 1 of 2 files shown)


libsys_airflow/plugins/data_exports/marc/oclc.py
```diff
@@ -43,6 +43,8 @@ def get_record_id(record: pymarc.Record) -> list:
                     oclc_number = raw_oclc_number[2:]
                 else:
                     oclc_number = raw_oclc_number
+                if oclc_number.startswith("0"):
+                    oclc_number = oclc_number[1:]
```
shelleydoljack, line 47: I'm not sure this is the correct approach. Could an "on" or "ocn" number legitimately start with 0? Maybe we should get the length of the digits when the string starts with a prefix and decide how much to strip off from there?

jermnelson: I'm not sure; this would be more of a question for @ahafele or Jeanette. I was thinking about this PR last night and I'm wondering if we have early records with multiple padding zeros that this approach wouldn't handle. Maybe we should use a regex instead to extract the oclc_number from none to n leading zeros.

ahafele: I don't think OCLC numbers with prefix will legitimately start with zero, based on this documentation. https://help.oclc.org/Metadata_Services/WorldShare_Collection_Manager/Data_sync_collections/Prepare_your_data/30035_field_and_OCLC_control_numbers

jermnelson: @shelleydoljack, I just commit a change that uses a regular expression to extract the OCLC number if it has 0 to n leading numbers.

shelleydoljack (approved):

PR #1155, opened 2024-08-26
Order data_export_marc view by HRID

jgreben:
This should hopefully fix the duplicated in the export.

### DIFF
--- a/libsys_airflow/plugins/data_exports/marc/exporter.py
+++ b/libsys_airflow/plugins/data_exports/marc/exporter.py
@@ -120,7 +120,7 @@ class Exporter(object):
         for row in instance_ids:
             marc_json_handler = marcJson()
             try:
-                marc_json_handler.elements(row[1])
+                marc_json_handler.elements(row[2])
                 marc21 = marc_json_handler.records[0]
             except Exception as e:
                 logger.warning(e)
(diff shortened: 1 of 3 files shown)


libsys_airflow/plugins/data_exports/marc/exporter.py
```diff
@@ -120,7 +120,7 @@ def retrieve_marc_for_full_dump(self, marc_filename: str, instance_ids: str) ->
         for row in instance_ids:
             marc_json_handler = marcJson()
             try:
-                marc_json_handler.elements(row[1])
+                marc_json_handler.elements(row[2])
                 marc21 = marc_json_handler.records[0]
```
jgreben, line 124: @jermnelson do you think we need to change the index of `marc_json_handler.records[0]` for any reason? I am not sure about the reference made by this index.

jermnelson: I don't think so as we're creating a new instance of the `marcJson` during each iteration of the `for` loop on line 121 so we always have one record in the `marc_jason_handler.records` list.

jermnelson (approved):

shelleydoljack:
🤞 this fixes it

Issue #1150, opened 2024-08-22
Use _dag_run_url function in oclc_reports in emails


jermnelson:
The `_dag_run_url` function in the oclc_reports.py module could be used in the email instead of what we're doing [here](https://github.com/sul-dlss/libsys-airflow/blob/34d24651d44f0848b2f9b575813fbbb9f190416c/libsys_airflow/plugins/data_exports/email.py#L155).

PR #1156, opened 2024-08-26
Refactoring of dag_run_url for email reports

### LINKED ISSUE #1143: Fix bug in calling webserver base_url
https://github.com/sul-dlss/libsys-airflow/blob/91a6ae8bd4f6d570611a52f4d932e5b0c5ced36f/libsys_airflow/plugins/data_exports/email.py#L192

airflow webserver base_url is a configuration not an airflow Variable, so needs to be gotten differently, like this: https://github.com/sul-dlss/libsys-airflow/blob/b78ceba6a560849006528a4086ab838495fa18a2/libsys_airflow/plugins/data_exports/email.py#L69

shelleydoljack:
Close #1143 and #1150

### DIFF
--- a/libsys_airflow/plugins/data_exports/email.py
+++ b/libsys_airflow/plugins/data_exports/email.py
@@ -1,5 +1,6 @@
 import logging
 import pathlib
+import urllib
 
 from jinja2 import Template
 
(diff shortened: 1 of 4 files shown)


shelleydoljack:
@jermnelson Maybe we can pair on fixing the mock for this in the test_oclc_reports.py

libsys_airflow/plugins/data_exports/email.py
```diff
@@ -42,6 +43,14 @@ def _cohort_emails():
     }
 
 
+def _dag_run_url(dag_run) -> str:
```
jermnelson, line 46: Just a minor, non-blocking request. In Python the convention is that private to the module functions start with an underscore `_`, since `_dag_run_url` is used in multiple modules, maybe rename this function to `dag_run_url`?

jermnelson (approved):
With one non-blocking request.

PR #1158, opened 2024-08-27
Change id seed in Interface Edit Javascript

### LINKED ISSUE #1077: VMA - Interfaces only allow for 2 MARC modifications
When adding an additional rule to move fields for example, one of the existing disappears. 
https://sul-libsys-airflow-stage.stanford.edu/vendor_management/interfaces/124

There was an original requirement during VMA development for Amaliv to have fields 505 and 520 moved to 905 and 920 due to enhanced content from vendor that may not be exported. This requirement is satisfied by interface MARC rules functionality but needs to be able to handle more than 2 rules in one one interface.

jermnelson:
Fixes #1077

When the existing remove or replace JSON was being rendered in the template, the id was being duplicated because the epoch int was returning the same value for multiple function calls (the server is too quick!)

### DIFF
--- a/libsys_airflow/plugins/vendor_app/templates/vendors/interface-edit.html
+++ b/libsys_airflow/plugins/vendor_app/templates/vendors/interface-edit.html
@@ -200,16 +200,17 @@
  function newRemoveFieldRule(marcField='') {
    const template = document.getElementById("remove-fields-template");
    const clone = template.content.cloneNode(true);
-   const now = new Date().getTime();
+   const counter =  document.querySelectorAll('div.remove-field').length + 1;
 
-   updateTemplate(clone, "remove-rule", now)
-   updateTemplateRemoveBtn(clone, "remove-rule-remove", "remove-rule", now)
-   updateInputTemplate(clone, "remove-field", now, marcField)
-   updateLabelTemplate(clone, "remove-field", now)
+   updateTemplate(clone, "remove-rule", counter)
+   updateTemplateRemoveBtn(clone, "remove-rule-remove", "remove-rule", counter)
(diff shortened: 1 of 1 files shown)


libsys_airflow/plugins/vendor_app/templates/vendors/interface-edit.html
```diff
@@ -200,39 +200,40 @@ <h2>MARC Processing Rules</h2>
...
-   updateLabelTemplate(clone, "move-indicator1-from", now)
-   updateInputTemplate(clone, "move-indicator2-from", now, fromIndicator2)
-   updateLabelTemplate(clone, "move-indicator2-from", now)
-   updateInputTemplate(clone, "move-field-to", now, toTag)
-   updateLabelTemplate(clone, "move-field-to", now)
-   updateInputTemplate(clone, "move-indicator1-to", now, toIndicator1)
-   updateLabelTemplate(clone, "move-indicator1-to", now)
-   updateInputTemplate(clone, "move-indicator2-to", now, toIndicator2)
-   updateLabelTemplate(clone, "move-indicator2-to", now)
+   const counter =  document.querySelectorAll('div.move-field').length + 1;
```
shelleydoljack, line 221: If there is not a `div.move-field` does querySelectorAll return a default null value? Or is there always a `div.move-field` element on the rendered page?

jgreben (approved):

libsys_airflow/plugins/vendor_app/templates/vendors/interface-edit.html
jermnelson: If there isn't a `div.move-field` input field the  `document.querySelectorAll('div.move-field').length` returns a value of 0.

PR #1160, opened 2024-08-27
Schedule pod for later in the evening

jgreben:
In case there are any enterprising metadata librarians without dinner plans...

If we run the dags too early in the evening and records are changed between 7pm and midnight we will end up skipping the records. This new schedule will run the pod and gobi dags at 11pm and and transmission dag at 2am.

### DIFF
--- a/libsys_airflow/dags/data_exports/default_schedules.json
+++ b/libsys_airflow/dags/data_exports/default_schedules.json
@@ -1,7 +1,7 @@
 {
   "schedule_gobi_days": 7,
-  "schedule_gobi_hours": 3,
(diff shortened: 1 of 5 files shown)


dlrueda:
We are setting ourselves up to miss updates even if we run this at 11pm.  What about if the run is on day x, we are selecting records up to the previous day? Then we are a day behind but we catch everything?

Even if we switch to the day before, it's ok if we ship this to run at 11 and deal with changing the selection later.  Let's get this shipped.

jgreben (commented):
I will submit a follow-up PR for any changes needed.

PR #1161, opened 2024-08-28
Change schedules to cron format

jgreben:
- Change schedules to cron format 
- rename default selection schedule variables.
- daily selection to/from dates changed to "yesterday" until "today"
- weekly selection to/from dates changed to "8 days ago" until "today"


jermnelson (approved):

libsys_airflow/dags/data_exports/default_schedules.json
shelleydoljack, line 1: I kind of wonder why we have this json file if we need to submit PR's to change the schedule anyways :shrug: . You could input the cron schedule like the payments DAG (including a timezone field to make timezone math "easier") https://github.com/sul-dlss/libsys-airflow/blob/7adcd3c41d787ef3644e1ad07387e47c13cb320a/libsys_airflow/dags/orafin/payments_to_orafin.py#L39-L41.

shelleydoljack (approved):

libsys_airflow/dags/data_exports/default_schedules.json
jgreben: My thought is that these PRs are to get it set up initially (since we are not quite primetime yet) and the variables will be to override the schedules temporarily as needed going forward without having to make PRs.

PR #1164, opened 2024-08-29
Replace custom FOLIO Client with FOLIO-FSE FOLIO Client

### LINKED ISSUE #801: Update FOLIO Client to use Refresh Token Rotation
Coming in Poppy and Quesnelia releases, Refresh Token Rotation (RTR) will be available with the current non-expiring token approach being depreciated with the non-expiring token being totally removed in the  Ramsons release. 

We will need to update libsys-airflow [client](https://github.com/sul-dlss/libsys-airflow/blob/main/libsys_airflow/plugins/folio/folio_client.py) to support RTR. The Ebsco FolioClient already uses RTR and can be used as a [pattern](https://github.com/FOLIO-FSE/FolioClient/blob/c8a850c644de3cf54c13f6fca9c53a1e8b9b353a/folioclient/FolioClient.py#L202-L229) to update the libsys-airflow client.

This [wiki page](https://wiki.folio.org/pages/viewpage.action?pageId=96414255) on the FOLIO wiki goes into more details.

jermnelson:
Fixes #801

TODO:
- [x] Fix failing tests
- [x] Test on folio-dev
  - [x] Data exports 
  - [x] VMA
  - [x] Orafin


shelleydoljack:
I _think_ we can start digital bookplates and use ebsco's FolioClient while this PR hangs out. Although we will be touching the orafin DAGs for digital bookplates so I'm not sure if we want to or should complete the testing for this PR at the beginning onf the workcycle.

jermnelson:
With @ahafele testing that hasn't revealed any issues with using EBSCO's folio_client, I think this PR can be merged.

shelleydoljack (approved):

PR #1170, opened 2024-09-03
Returns 404 Error Page for missing Vendors, Interfaces, and Files

### LINKED ISSUE #1084: Return 404 Errors for Missing VMA Database Objects
We're getting a small but steady stream of [HoneyBadger](https://app.honeybadger.io/projects/111133/faults/100338623) errors when a user attempts to access a `Vendor`, `VendorInterface`,  or `VendorFile` through the user interface for a database object that doesn't currently exist. This may be because the original database object was deleted or never existed. 

Instead of generating a HB alert, we should just return a custom 404 message that the database object doesn't exist.

jermnelson:
Fixes #1084

### DIFF
--- a/libsys_airflow/plugins/vendor_app/vendor_management.py
+++ b/libsys_airflow/plugins/vendor_app/vendor_management.py
@@ -6,7 +6,7 @@ from datetime import datetime
 from airflow.api.common.trigger_dag import trigger_dag
 from airflow.models import Variable
 from flask_appbuilder import expose, BaseView
-from flask import request, redirect, url_for, flash, send_from_directory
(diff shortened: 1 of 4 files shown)


shelleydoljack (commented):
Will airflow show some "404 not found" page when the status code is returned? Or do you need to add a response body?

jermnelson:
> Will airflow show some "404 not found" page when the status code is returned? Or do you need to add a response body?

Yes, Airflow will just display the following screen if the database object isn't found:
Screenshot 2024-09-03 at 2 54 44 PM

jgreben (approved):

shelleydoljack (approved):

Issue #1180, opened 2024-09-05
Add or update digital bookplate data in bookplate table


shelleydoljack:
This task gets the parsed JSON bookplate data and store it in the bookplates table. It should lookup in the bookplates table for the Symphony Fund Name (e.g. ABBOTT) and druid (or just druid for the "gift funds") to see if a record already exists. If none found, create a new record in the table, with created date and updated date the same. If a record is found, compare the druid, titles and image filenames in the record to see if there are changes. If no changes, do nothing. If there are changes, update the data in the bookplates table (also updating the updated date column) and pass the updated data along with a reason field (e.g. title change, filename change) in the XCOM so that it can be reported out. For new bookplates, add that to the XCOM too to be reported out.

shelleydoljack:
A database conflict should tell you which field it errored on so we should be able to catch that exception to pass in the XCOM. We want to report out if the filename or title changed if it's not too difficult but it doesn't happen often so we might add this later.

PR #1193, opened 2024-09-09
Adds email report for new and updated digital bookplates.

### LINKED ISSUE #1184: Send email report when bookplate metadata changed or is new
This task should receive data from the XCOM for the task outlined in #1180. Updated bookplate metadata and new bookplate metadata should be emailed to sul-unicorn-devs list and linchang@stanford.edu. Only email to sul-unicorn-devs from non-prod server.

The email header could be "Digital bookplates new and updated" and the body should contain all fields from the bookplates table (tab-separated), such as:

**New digital bookplates data**
Fund name\tab123cd4567\timage filename\tTitle

**Updated digital bookplates data**
Fund name\tab123cd4567\timage filename\tTitle

shelleydoljack:
Closes #1184


libsys_airflow/plugins/digital_bookplates/email.py
```diff
@@ -0,0 +1,94 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def _new_updated_bookplates_email_body(new: list, updated: list):
+    template = Template(
+        """
+        <h2>New digital bookplates metadata</h2>
+        {% if new|length > 0 %}
+        <ul>
```
jermnelson, line 19: Would creating a table instead of unordered list be easier to understand in the email? Each line would be a table row so that we're not depending on `\t` to separate each value in the line? i.e
```
<tr>
  <th>Fund Name</th>
  <th>Druid</td>
  <th>Bookplate Image</th>
  <th>Reason</th>
 </tr>
 ...
<tr>
   <td>Fund name</td>
   <td>ab123cd4567</td>
    <td>image_ab123cd4567.jp2</td>
    <td>Title</td>
</tr>
```

shelleydoljack: 🤔 I see you added Reason as a header. We should add reason to tracking changes I think. I will update #1180 to add "reason" for an update. A table in the email would definitely make it easier to read. So I'll make that change. 
I think this task will be getting data from the XCOM from the task that adds or updates our table data. I'm not sure what our structure in the XCOM will look like exactly. I'll change the fixture to something more JSON-ny than a tsv string. Thanks for the suggestion!

jgreben (changes requested):
I don't think the OKAPI_URL variable needs to be in the test fixture.

tests/digital_bookplates/test_bookplates_emails.py
```diff
@@ -0,0 +1,146 @@
...
+            case "FOLIO_URL":
+                value = "folio-test"
+
+            case "EMAIL_DEVS":
+                value = "test@example.com"
+
+            case "BOOKPLATES_EMAIL":
+                value = "nobody@example.com"
+
+            case "OKAPI_URL":
```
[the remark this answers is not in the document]
shelleydoljack: Yes, because the digital bookplates email calls the is_production function.

jgreben (approved):

jermnelson (approved):

PR #1195, opened 2024-09-10
New SQLAchemy Digital Bookplate Model

### LINKED ISSUE #1175: Create a digital bookplate model
Similar (but not as complicated) to the [VMA models](https://github.com/sul-dlss/libsys-airflow/blob/main/libsys_airflow/plugins/vendor/models.py), we should create a table to store digital bookplate metadata, containing the fields we need to create 979s: fund name, image filename, druid, and title. We use the fund name as primary key. If there is no fund name (there are plenty of these), we use the druid as primary key. We will use this table to update data we get from Argo, report out if there are changes (new bookplates, deleted bookplates, different titles or image filenames). We will also extract the data from this table to add 979 fields in folio MARC records and display in a UI for users to select which fund name (or "no fund" druid) to add 979 fields.

Previously, we stored this data in a tsv, which can be found on bodoni at `/s/SUL/Config/BookplateMerge/symp_dor.txt`.

### LINKED ISSUE #1182: Add created and updated date columns for bookplates table
We probably should add created date and updated date columns in the bookplates table in order to keep track of when data might go missing.

jermnelson:
Fixes #1175
Fixes #1182 

TODO:
- [x] Enable Alembic migrations 
- [x] Add DB init to Docker Compose yaml files


config/deploy.rb
```diff
@@ -75,21 +77,21 @@
   desc 'Run Alembic database migrations'
```
jermnelson, line 77: @shelleydoljack and @jgreben would it make sense to add a parameter to these alembic to pass in the `name` as parameter to these tasks i.e. `cap dev alembic --name digital_bookplates` or something similar?

jgreben: IMO no, because there is the `WHERE NOT EXISTS` clause, and If or when someone is spinning up a new system (locally or otherwise) they would presumably want all of the databases created without having to manually run the command. Maybe they can be added them as the PSQL_ARGS or something where it will run the command in for each? I am not sure how that works.

jgreben (approved):

PR #1217, opened 2024-09-12
Retrieve and process Digital Bookplate Metadata

### LINKED ISSUE #1177: Get digital bookplate metadata JSON documents
Blocked by #1175 
For each member druid of the digital bookplate collection, lookup the public JSON document at `purl.stanford.edu` and parse out the fields we will store in our bookplates table. This [Bookplates class](https://github.com/sul-dlss/BookplateMerge/blob/main/bookplate.rb) from our old process should be used as a guide to the fields we need from the JSON. The output should be sending parsed data to a task that will store data in the bookplates table. If the public JSON document does not contain all the fields necessary for us to create 979's (like missing image filename for instance), pass this along via XCOM to be reported in an email.

[Example JSON](https://purl.stanford.edu/ws066yy0421.json) for the ABBOTT bookplate. The fields we need:
*fund name*
`description.identifier.value` where displayLabel="Symphony Fund Name"
```
"identifier": [
      {
        "structuredValue": [],
        "parallelValue": [],
        "groupedValue": [],
        "value": "1065089-106-KAUNB",
        "identifier": [],
        "displayLabel": "Symphony Fund ID",
        "note": [],
        "appliesTo": []
      },
      {
        "structuredValue": [],
        "parallelValue": [],
        "groupedValue": [],
        "value": "ABBOTT",
        "identifier": [],
        "displayLabel": "Symphony Fund Name",
        "note": [],
        "appliesTo": []
      },
      {
        "structuredValue": [],
        "parallelValue": [],
        "groupedValue": [],
        "value": "KAUNB",
        "identifier": [],
        "displayLabel": "Fund Number",
        "note": [],
        "appliesTo": []
      },
      {
        "structuredValue": [],
        "parallelValue": [],
        "groupedValue": [],
        "value": "ws066yy0421_00_0001.jp2",
        "identifier": [],
        "displayLabel": "Image filename",
        "note": [],
        "appliesTo": []
      }
    ]
```
*druid*
`externalIdentifier` field:
`"externalIdentifier": "druid:ws066yy0421",`
*image filename*
This appears in the `description.identifiers` list and the `structural.contains` list. Per Andrew, 

> In the JSON, the description will have all the fields that map to MODS and the structural will have the fields that are in contentMetadata.

From our previous process, we used the `<contentMetadata>` to get to the image filename. We should pull from the JSON equivalent, so from the `structural.contains.structural.contains` list:
```
"structural": {
    "contains": [
      {
        "type": "https://cocina.sul.stanford.edu/models/resources/image",
        "externalIdentifier": "https://cocina.sul.stanford.edu/fileSet/ws066yy0421-ws066yy0421_1",
        "label": "Image 1",
        "version": 5,
        "structural": {
          "contains": [
            {
              "type": "https://cocina.sul.stanford.edu/models/file",
              "externalIdentifier": "https://cocina.sul.stanford.edu/file/ws066yy0421-ws066yy0421_1/ws066yy0421_00_0001.jp2",
              "label": "ws066yy0421_00_0001.jp2",
              "filename": "ws066yy0421_00_0001.jp2",
              "size": 297491,
              "version": 5,
              "hasMimeType": "image/jp2",
              "sdrGeneratedText": false,
              "correctedForAccessibility": false,
              "hasMessageDigests": [
                {
                  "type": "sha1",
                  "digest": "ff56bbe8c4de850b7d1111cafe60b527999f5fea"
                },
                {
                  "type": "md5",
                  "digest": "fc30918e85556b0a2b6ee2a77d46c8d1"
                }
              ],
              "access": {
                "view": "world",
                "download": "world",
                "controlledDigitalLending": false
              },
              "administrative": {
                "publish": true,
                "sdrPreserve": false,
                "shelve": true
              },
              "presentation": {
                "height": 1392,
                "width": 1081
              }
            }
          ]
        }
      }
    ],
    "hasMemberOrders": [],
    "isMemberOf": [
      "druid:nh525xs4538"
    ]
  }
```
I think we will always take the first in those list. I will confirm.

jermnelson:
Fixes #1177

In this approach, each purl URL is in a separate task that retrieves the Cocina and extracts the title (label), image filename, and fund name. I initially throttle the number of active DAG runs to 5 and retrieving and extracting 817 druids took 22:25 minutes. I will do a second run with active DAG runs set to 10 to see if that improves the performance.


libsys_airflow/plugins/digital_bookplates/purl_fetcher.py
```diff
@@ -1,6 +1,13 @@
 import httpx
 
 from airflow.decorators import task
+from jsonpath_ng.ext import parse
+
+FUND_NAME_EXPR = parse(
+    "$.description.identifier[?(@.displayLabel == 'Symphony Fund Name')].value"
+)
+IMAGE_FILE_NAME_EXPR = parse("structural.contains[0].structural.contains[0].filename")
+TITLE_EXPR = parse("$.label")
```
shelleydoljack, line 10: Do you know if the label field updates if the `structuredValue.value` fields in `description.title` changes? It looks to be the case for https://purl.stanford.edu/ws066yy0421.json (and definitely easier to parse!).

libsys_airflow/plugins/digital_bookplates/purl_fetcher.py
```diff
@@ -16,10 +23,47 @@ def fetch_druids() -> list:
...
+    druid = druid_url.split("/")[-1].split(".json")[0]
+
+    metadata = {
+        "druid": druid,
+        "image_file": None,
+        "failure": None,
+        "fund_name": None,
+        "title": None,
+    }
+    try:
```
shelleydoljack, line 38: I kind of wondered if we would define models using attrs like with [orafin](https://github.com/sul-dlss/libsys-airflow/blob/main/libsys_airflow/plugins/orafin/models.py) but maybe not since we write the data to a table.

shelleydoljack (approved):
👍

libsys_airflow/plugins/digital_bookplates/purl_fetcher.py
jermnelson: I think the SQLAlchemy provides the equivalent functionality we used for the Orafin attrs models.

libsys_airflow/plugins/digital_bookplates/purl_fetcher.py
jermnelson: I think so? This might be better answered by one of the experienced SDR developers or Andrew.

andrewjbtw: I think it is (derived from `description` data) but that field might be deprecated in the future. The plain "label" was a Fedora field and carried over for some backwards compatibility reasons. There isn't a consensus on whether to keep the field or not.

shelleydoljack: @andrewjbtw Do you know if other consumers use the label field or would our digital bookplates stuff be the only thing (and likely forgotten to and then it goes away and we have to update our stuff to get from the description object)?

andrewjbtw: It would be one of the few consumers. If I had to guess, these are the two likely scenarios for that field: 

- devs prefer to use it for ease of parsing (vs. all the ways to structure `title`), and it actually gets used more
- it gets deprecated

Currently, I think most Purl consuming applications still parse the MODS title (not label and not Cocina - yet). But we discovered during the Fedora migration that there were a few places where an application parsed label instead. I don't recall which cases off the top of my head but the situation where something broke because of a change to the label field has happened a couple of times.

shelleydoljack: @jermnelson I think we should just keep using label until we can't. Feel free to merge or change what we do.

PR #1219, opened 2024-09-14
Adds/Updates Digital Bookplates based on Druid JSON

### LINKED ISSUE #1180: Add or update digital bookplate data in bookplate table

jermnelson:
Fixes #1180


tests/digital_bookplates/test_purl_fetcher.py
```diff
@@ -1,13 +1,50 @@
...
+        image_filename="gc698jf6425_00_0001.jp2",
+        fund_name="RHOADES",
+        title="John Skylstead and Carmel Cole Rhoades Fund for California History and the History of the North American West",
+    ),
+    DigitalBookplate(
+        id=3,
+        created=datetime.datetime(2024, 9, 13, 17, 16, 15, 986798),
+        updated=datetime.datetime(2024, 9, 13, 17, 16, 15, 986798),
+        druid="ab123xy4567",
+        fund_name=None,
```
shelleydoljack, line 41: I keep wanting to see where we use the druid when there is not a fund name, but that's just how we handled it previously since we were using fund name as a key from the tsv file. I have to remember that since we are saving the data to a table, we don't need to use fund name as a key in a perl hash.

shelleydoljack (approved):

PR #1225, opened 2024-09-17
Handle Deleted from Argo Druids

### LINKED ISSUE #1181: Track and report out when fund name is not in Argo anymore
This might have happened once or twice over the years: a fund name was in Argo the previous week but on the next week's run, it wasn't there. This was reported out in an email. We could compare all the druids in our bookplates table against the list of druids that we got from fetching the member druids from purl-fetcher in order to figure out what fund name is missing from Argo. Or we can compare all the fund names (primary key) from the bookplates table with a list of fund names extracted from the XML documents and report out when it exists in the table but not the list. When this happens, we should create another column in the bookplates table for "deleted from Argo" rather than deleting the record in the bookplates table.

jermnelson:
Fixes #1181 

Also adds new boolean Column to DigitalBookplates model set when a Digital Bookplate Druid is removed from Argo.


jgreben (approved):
Just wondering about the PR title ;)

PR #1230, opened 2024-09-18
Task to return list of uuids for folio funds 

### LINKED ISSUE #1208: Task that looks up fund UUIDs in folio
The output of this task is to provide a list of fund UUIDs to check against invoice lines paid to determine if it was paid using a bookplate fund. We could get all funds from folio and compare the name to fund_name in our bookplates table, or we can get all the fund_names in our bookplates table and query folio for the fund UUID.

jgreben:
Returns funds matching digital_bookplates table funds.

Fixes #1208

### DIFF
--- a/libsys_airflow/plugins/digital_bookplates/bookplates.py
+++ b/libsys_airflow/plugins/digital_bookplates/bookplates.py
@@ -1,19 +1,51 @@
 import logging
 
 from airflow.decorators import task
+from airflow.models import Variable
+from airflow.providers.postgres.hooks.postgres import PostgresHook
+from sqlalchemy.orm import Session
+from libsys_airflow.plugins.digital_bookplates.models import DigitalBookplate
+from folioclient import FolioClient
 
 logger = logging.getLogger(__name__)
 
 
+def _folio_client():
+    return FolioClient(
+        Variable.get("OKAPI_URL"),
+        "sul",
+        Variable.get("FOLIO_USER"),
+        Variable.get("FOLIO_PASSWORD"),
+    )
+
+
 @task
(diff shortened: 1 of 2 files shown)


jermnelson (approved):

shelleydoljack (changes requested):
Apologies I didn't update the ticket earlier. We need this to return a dictionay instead of a list. The dictionary should be something like `{"druid": "fund_uuid"}`. When we find instances paid on bookplate funds, we will only have the fund UUID. We need a way to lookup in our bookplates table the data to pass to the next DAG to add 979s. We should be able to pull from the XCOM this dictionary during the DAG run.

libsys_airflow/plugins/digital_bookplates/bookplates.py
```diff
@@ -1,19 +1,53 @@
...
     Looks up in bookplates table for fund_name
     Queries folio for fund_name
-    Returns list of fund UUIDs
+    Returns dict of fund UUIDs
     """
-    # {"druid": "fund_uuid"}
+    folio_client = _folio_client()
+    folio_funds = folio_client.folio_get(
+        "/finance-storage/funds", query_params={"limit": 2999}
+    )
```
shelleydoljack, line 32: FYI you could add `key="funds"` to the `folio_get` request and it will return a list of fund objects, e.g. `[{'id': '8527a67d-f099-4a19-b104-7cd8969569b1', etc.}, {'id': etc.}]` instead of `{'funds':[<list of fund objects]}`.

shelleydoljack (approved):
Thank you!

PR #1240, opened 2024-10-03
Change to use underscore instead of dash for filename

### LINKED ISSUE #1237: VMA file name question
There is a FOLIO bug that is causing some issues with files sent from Data Import - see write up [here](https://folio-org.atlassian.net/browse/MODDATAIMP-1094). 

TLDR - when a file name is a string of numbers followed by - the numbers get cut off leaving all VMA file names looking like e.g. **processed.mrc** in the data import logs. What I am wondering is, how hard would it be for the VMA to switch from using `-` to `_`? The underscore does not cause this same issue. 

Screenshot 2024-10-02 at 10 34 16 AM

jermnelson:
Fixes #1237

### DIFF
--- a/libsys_airflow/plugins/vendor/marc.py
+++ b/libsys_airflow/plugins/vendor/marc.py
@@ -133,7 +133,7 @@ def process_marc(
                     _add_fields(record, add_fields)
                 records.append(record)
 
-    new_marc_path = marc_path.with_stem(f"{marc_path.stem}-processed")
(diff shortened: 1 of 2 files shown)


shelleydoljack (commented):
Do you know if this will affect how we check if we've downloaded and processed the marc file? Are we basing it off of `marc_path.stem`? I'm not very familiar with how VMA works. Maybe it won't be a problem b/c `-processed` or `_processed` is not taken into account at all when querying the vendor files table? 🤷

shelleydoljack (approved):

PR #1244, opened 2024-10-03
 Organizes Fetch Digital Bookplates DAG email into a group

jermnelson:
Now looks like:
Screenshot 2024-10-03 at 12 29 45 PM


shelleydoljack (approved):
😎

PR #1245, opened 2024-10-04
Triggers Digital Bookplates Instances DAG

### LINKED ISSUE #1221: Trigger digital bookplate instances DAG when there is new bookplates data
When there are new bookplates identified from the fetch_digital_bookplates DAG, we want to automatically trigger the DAG that finds instances paid on bookplate funds. The data passed to trigger the dag should be a list of new funds (bookplate metadata structure) and a date of beginning of time on FOLIO (date should be 8/28/2023; there are 3 paid invoices prior to 9/1/2023).

When the DAG has this configuration:
```
@dag(
    default_args=default_args,
    schedule=CronDataIntervalTimetable(
        cron="0 2 * * WED", timezone="America/Los_Angeles"
    ),
    start_date=datetime(2023, 8, 28),
    catchup=False,
    tags=["digital bookplates"],
)
def digital_bookplate_instances():
```
And it is triggered w/out a config, it creates dag_runs that have `data_interval_end` and `data_interval_start` dates we can use on a weekly basis:
```
{
  "dag_runs": [
    {
      "conf": {},
      "dag_id": "digital_bookplate_instances",
      "dag_run_id": "manual__2024-09-19T20:23:18.515617+00:00",
      "data_interval_end": "2024-09-18T09:00:00+00:00",
      "data_interval_start": "2024-09-11T09:00:00+00:00",
      "end_date": null,
      "execution_date": "2024-09-19T20:23:18.515617+00:00",
      "external_trigger": true,
      "last_scheduling_decision": "2024-09-19T20:23:31.696585+00:00",
      "logical_date": "2024-09-19T20:23:18.515617+00:00",
      "note": null,
      "run_type": "manual",
      "start_date": "2024-09-19T20:23:18.601494+00:00",
      "state": "running"
    },
    {
      "conf": {},
      "dag_id": "digital_bookplate_instances",
      "dag_run_id": "scheduled__2024-09-11T09:00:00+00:00",
      "data_interval_end": "2024-09-18T09:00:00+00:00",
      "data_interval_start": "2024-09-11T09:00:00+00:00",
      "end_date": null,
      "execution_date": "2024-09-11T09:00:00+00:00",
      "external_trigger": false,
      "last_scheduling_decision": "2024-09-19T20:23:31.693525+00:00",
      "logical_date": "2024-09-11T09:00:00+00:00",
      "note": null,
      "run_type": "scheduled",
      "start_date": "2024-09-19T20:23:18.601346+00:00",
      "state": "running"
    }
  ],
  "total_entries": 2
}
```
But for dag runs for new bookplates, we need to trigger the dag with a configuration. When triggering the above defined dag with a logical_date value of 8/28/2023 00:00:00, the dag run that gets created is this:
```
{
      "conf": {},
      "dag_id": "digital_bookplate_instances",
      "dag_run_id": "manual__2023-08-28T00:00:00+00:00",
      "data_interval_end": "2023-08-23T09:00:00+00:00",
      "data_interval_start": "2023-08-16T09:00:00+00:00",
      "end_date": null,
      "execution_date": "2023-08-28T00:00:00+00:00",
      "external_trigger": true,
      "last_scheduling_decision": "2024-09-19T20:26:19.583010+00:00",
      "logical_date": "2023-08-28T00:00:00+00:00",
      "note": null,
      "run_type": "manual",
      "start_date": "2024-09-19T20:26:10.373003+00:00",
      "state": "running"
    }
```
We should use the logical_date field (execution_date is deprecated (See [Release Notes Miscellaneous](https://airflow.apache.org/docs/apache-airflow/stable/release_notes.html#id15))) when the run_id starts with "manual_" for the invoice query params `paymentDate>=logical_date`

jermnelson:
Fixes #1221 
New task In the `fetch_digital_bookplates` triggers  `digital_bookplate_instances` when new funds are added.

Screenshot 2024-10-04 at 12 09 34 PM

In the `digital_bookplate_instances` these new funds are retrieved from the context in a new `get_funds` task:

Screenshot 2024-10-04 at 12 10 01 PM

### DIFF
--- a/libsys_airflow/plugins/folio/invoices.py
+++ b/libsys_airflow/plugins/folio/invoices.py
@@ -142,7 +142,7 @@ def invoice_lines_from_invoices(invoices: list) -> list:
 
 
 @task
-def filter_invoice_lines(invoice_lines: list) -> list:
+def filter_invoice_lines(invoice_lines: list, funds: list) -> list:
     """
     Given a list of invoice line dictionaries,
     Filters invoice lines to only those with fund IDs and po line IDs
--- a/libsys_airflow/dags/digital_bookplates/digital_bookplate_instances.py
+++ b/libsys_airflow/dags/digital_bookplates/digital_bookplate_instances.py
@@ -1,8 +1,10 @@
 from datetime import datetime, timedelta
 
-from airflow.decorators import dag
+from airflow.decorators import dag, task
 from airflow.operators.empty import EmptyOperator
+from airflow.operators.python import get_current_context
 from airflow.timetables.interval import CronDataIntervalTimetable
+
 
 from libsys_airflow.plugins.digital_bookplates.bookplates import (
     bookplate_fund_po_lines,
@@ -43,11 +45,19 @@ def digital_bookplate_instances():
 
     end = EmptyOperator(task_id="end")
 
+    @task
+    def get_funds() -> list:
+        context = get_current_context()
+        params = context.get("params", {})  # type: ignore
+        return params.get("funds", [])
+
+    funds = get_funds()
+
     retrieve_paid_invoices = invoices_paid_within_date_range()
 
     retrieve_paid_invoice_lines = invoice_lines_from_invoices(retrieve_paid_invoices)
 
-    filter_paid_invoice_lines = filter_invoice_lines(retrieve_paid_invoice_lines)
+    filter_paid_invoice_lines = filter_invoice_lines(retrieve_paid_invoice_lines, funds)
 
     retrieve_bookplate_fund_po_lines = bookplate_fund_po_lines(
         filter_paid_invoice_lines
@@ -57,7 +67,7 @@ def digital_bookplate_instances():
 
     launch_add_tag_dag = launch_add_979_fields_task(instances=retrieve_instances)
 
-    start >> retrieve_paid_invoices
+    start >> [funds, retrieve_paid_invoices]
     launch_add_tag_dag >> end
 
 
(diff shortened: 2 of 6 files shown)


libsys_airflow/plugins/folio/invoices.py
```diff
@@ -142,7 +142,7 @@ def invoice_lines_from_invoices(invoices: list) -> list:
 
 
 @task
-def filter_invoice_lines(invoice_lines: list) -> list:
+def filter_invoice_lines(invoice_lines: list, funds: list) -> list:
```
shelleydoljack, line 145: I thought we decided to add the funds list to the `bookplate_fund_po_lines` task? I mean, we could do it in `filter_invoice_lines` but then it wouldn't really be a generic folio plugin task. Maybe we don't care to make it a generic folio task? I can't think of a use case where we would need a list of invoice lines and their related po lines and fund ids.... I picked up the ticket to do the bookplate_fund_po_lines task so I could refactor filter_invoice_lines and collapse it into bookplate_fund_po_lines or move the funds to bookplate_fund_po_lines. Thoughts?

shelleydoljack (changes requested):

libsys_airflow/plugins/folio/invoices.py
jermnelson: Good question, after looking over the DAG flow, I initial thought that if the resulting list of invoice lines with corresponding po_lines and fund_ids from `filter_invoice_lines` may include a lot of data that never gets used in downstream tasks, then doing the filtering here would mean we are processing only those Instances that we care about later in the DAG. Happy to change flow to use bookplate_fund_po_lines to filter the funds instead.

shelleydoljack: 👍 Let's cleanup/DRYup in a subsequent PR. I think merging this will cause some conflicts with my [PR](https://github.com/sul-dlss/libsys-airflow/pull/1246) but oh well. I'll fix next week.

shelleydoljack (approved):

PR #1246, opened 2024-10-04
Adds bookplate_fund_po_lines task and test.

### LINKED ISSUE #1243: create bookplate_fund_po_lines task
This task checks if fund Id from invoice lines data struct contains a bookplate fund. The input of this task is a data structure like this:
```
    [
      {
        "fadacf66-8813-4759-b4d3-7d506db38f48": {
          "fund_ids": [
            "0e8804ca-0190-4a98-a88d-83ae77a0f8e3"
          ],
          "poline_id": "b5ba6538-7e04-4be3-8a0e-c68306c355a2"
        }
      }
    ]
```
It is "invoice_line_uuid": with fund_ids and poline_id objects. This task needs to lookup in the digital bookplates table the fund_id and remove from the list objects that are not in the bookplates table. There could be mutliple fund_ids per invoice line. If at least 1 of them is a bookplate fund, we want to keep it for adding a 979 to the instance. We don't need to keep the invoice line in our output data at this point and we should use this to get the data we need to add 979 fields so the output should return po lines and the bookplate metadata, e.g.:
```
[
  {
    "bookplate_metadata": { "druid": "", "fund_name": "", "image_filename": "", "title": "" },
    "poline_id": [
      "798596da-12a6-4c6d-8d3a-3bb6c54cb2f1",
      "7a5888fe-689b-4cfe-a27d-c2675a235203"
    ]
  }
]
```

shelleydoljack:
Closes #1243

### DIFF
--- a/libsys_airflow/plugins/digital_bookplates/bookplates.py
+++ b/libsys_airflow/plugins/digital_bookplates/bookplates.py
@@ -19,6 +19,31 @@ def _folio_client():
     )
 
 
+def _get_bookplate_metadata_with_fund_uuids() -> dict:
+    funds = {}
+    pg_hook = PostgresHook("digital_bookplates")
+    with Session(pg_hook.get_sqlalchemy_engine()) as session:
+        for f in (
+            session.query(
+                DigitalBookplate.fund_uuid,
+                DigitalBookplate.fund_name,
+                DigitalBookplate.druid,
+                DigitalBookplate.image_filename,
+                DigitalBookplate.title,
+            )
+            .where(DigitalBookplate.fund_uuid.is_not(None))
+            .all()
+        ):
+            funds[f.fund_uuid] = {
+                "fund_name": f.fund_name,
+                "druid": f.druid,
+                "image_filename": f.image_filename,
+                "title": f.title,
+            }
+
+    return funds
+
+
 @task
 def bookplate_fund_ids(**kwargs) -> dict:
     """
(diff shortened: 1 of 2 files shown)


jgreben (approved):

libsys_airflow/plugins/digital_bookplates/bookplates.py
```diff
@@ -19,6 +19,31 @@ def _folio_client():
     )
 
 
+def _get_bookplate_metadata_with_fund_uuids() -> dict:
+    funds = {}
+    pg_hook = PostgresHook("digital_bookplates")
+    with Session(pg_hook.get_sqlalchemy_engine()) as session:
+        for f in (
+            session.query(
+                DigitalBookplate.fund_uuid,
```
jermnelson, line 28: I think this can be collapsed into 

```session.query(DigitalBoookplate).where(DigitalBookplate.fund_uuid.is_not(None))```

jermnelson (approved):
Approved; non-blocking suggestion.

libsys_airflow/plugins/digital_bookplates/bookplates.py
shelleydoljack: Good idea https://github.com/sul-dlss/libsys-airflow/pull/1246/commits/ed17fdf5ec353f1e468756f04e1bdd0608af893b

PR #1252, opened 2024-10-07
Query for instanceId from PO Line ID

### LINKED ISSUE #1210: Get instances paid on bookplate fund
The output of this task is a list of instance UUIDs and associated fund_name from the bookplates table. We get here by having a list of po lines that were linked to an invoice line that paid on a bookplate fund. The previous task on [ticket #1209](https://github.com/sul-dlss/libsys-airflow/issues/1209) should give us this data (fund name and po lines). For each po line, find out the `instanceId` from the `/orders/order-lines` endpoint. Not all po lines will have an `instanceId` field because the po line is linked to a package (see https://folio.stanford.edu/orders/lines/view/9047a74c-1057-4ddb-b52d-d01429b0b450?limit=50&offset=0&qindex=poLineNumber&query=55487-1 for example). Once we have the instance UUID and associated fund_name, this data will be used to trigger the add 979 DAG.

jermnelson:
Fixes #1210

### DIFF
--- a/tests/test_orders.py
+++ b/tests/test_orders.py
@@ -11,7 +11,7 @@ mock_order_lines = {
     "84701022-43af-401e-bcea-fb5fbd8f49f3": {
         "instanceId": "e6803f0b-ed22-48d7-9895-60bea6826e93"
     },
-    "9f7031df-d30b-40c2-955a-7d522c303a43": {"instanceId": None},
+    "9f7031df-d30b-40c2-955a-7d522c303a43": {},
 }
 
 
--- a/tests/test_orders.py
+++ b/tests/test_orders.py
@@ -1,0 +1,97 @@
+import pytest
+
+from unittest.mock import MagicMock
+
+from libsys_airflow.plugins.folio.orders import instances_from_po_lines
+
+mock_order_lines = {
+    "8c91eb1c-4e32-44e8-bcce-b34850bdf3d5": {
+        "instanceId": "e6803f0b-ed22-48d7-9895-60bea6826e93"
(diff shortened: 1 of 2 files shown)


tests/test_orders.py
```diff
@@ -0,0 +1,97 @@
...
+from libsys_airflow.plugins.folio.orders import instances_from_po_lines
+
+mock_order_lines = {
+    "8c91eb1c-4e32-44e8-bcce-b34850bdf3d5": {
+        "instanceId": "e6803f0b-ed22-48d7-9895-60bea6826e93"
+    },
+    "84701022-43af-401e-bcea-fb5fbd8f49f3": {
+        "instanceId": "e6803f0b-ed22-48d7-9895-60bea6826e93"
+    },
+    "9f7031df-d30b-40c2-955a-7d522c303a43": {"instanceId": None},
```
shelleydoljack, line 14: This mock isn't quite accurate. The key `instanceId` could be completely missing from the order-line response. I think that might change how the function works?

jermnelson: Ok, updated the fixture with an empty dictionary. Because we're doing a get to retrieve the `instanceId` here on [32](https://github.com/sul-dlss/libsys-airflow/blob/273c40046d4ee5d31bc1ac096e646bd93e37d52b/libsys_airflow/plugins/folio/orders.py#L32), a missing `instanceId` will return a None instead of throwing a KeyError.

jgreben (approved):

shelleydoljack (approved):
👍

Issue #1254, opened 2024-10-08
Create a table for digital bookplate batches


shelleydoljack:
The table will be used to keep track of when users trigger the add 979 field DAG manually through the UI. Previously, this table would contain the following information: submitted date, created by, report email, batch type (add or delete), bookplate druid, filename, ckey count, completed date.
![Screenshot 2024-10-08 at 11 24 36 AM](https://github.com/user-attachments/assets/993d942a-9038-4f1a-9d65-7f33db524288)

We also presented the user with batches in the queue (not yet processed). It showed the same information as the completed batches, driven by a query where completed_date is null. The bookplate batches table could have a separate "processed" or "completed" column if that would make presenting completed/in queue information easier.

shelleydoljack:
As an airflow admin user, I can see the status of the dag runs. As a user who has access to upload files to the UI, I do not have the ability to view the DAGs and the statuses of their runs. Creating a table to keep track of bookplate batches (and the views) is a "nice to have" and is not necessary for the work cycle. This could be prioritzed once more of the dags are completed.

shelleydoljack:
This table was used to figure out what needed to be processed when the script to add 979 fields ran on a schedule outside of business hours. Since the DAG to add 979 will run whenever it is triggered, then we don't need to keep track of queued/processed. I'm going to move this to the Icebox since it is not required for the work cylce.

Issue #1186, opened 2024-09-06
Create a DAG and task that starts the add 979 fields DAG


shelleydoljack:
The input for the add 979 fields DAG will be a list of instance UUIDs and the associated bookplate fund data. The task should get the bookplate data from the bookplates table and contruct a 979 tag with the fund name in subfield f, druid in subfield b, image filename in subfield c, and title in subfield d, such as:
`{'979': {'ind1': ' ', 'ind2': ' ', 'subfields': [{'f': 'ABBOTT'}, {'b': 'druid:ws066yy0421'}, {'c': 'ws066yy0421_00_0001.jp2'}, {'d': 'The The Donald P. Abbott Fund for Marine Invertebrates'}]}}`
The instance UUID and this data should be passed to the function to add a tag (described in #1185 ).

shelleydoljack:
I think it would be good to scaffold out this DAG to help define how we will trigger it (#1238). So this ticket should also include creating some rough outline of a DAG with the following steps:
1. lookup bookplate table for data to construct a 979 tag
2. get the MARC for each instance (see section [Get the MARC for each instance to do a PUT to add 979](https://github.com/shelleydoljack/notebook/blob/main/digital_bookplates/instances_paid_on_bookplate_funds.ipynb) for code examples)
3. pass data along to a function to add a tag #1185
4. catch errors in updating the MARC data to be emailed in a report
5. generate email report of adding 979 data

jermnelson:
The incoming data will be a dictionary made of the Instance UUID as the key and the value is a list of bookplate metadata. 

An example:
```
{
 "242c6000-8485-5fcd-9b5e-adb60788ca59": [
  { "druid": "", "fund_name": "", "image_filename": "", "title": "" },
  { "druid": "", "fund_name": "", "image_filename": "", "title": "" },
 ],
 "bd65eb13-739b-4aa8-baaa-9c1ea48f4c33": [
  { "druid": "", "fund_name": "", "image_filename": "", "title": "" },
  { "druid": "", "fund_name": "", "image_filename": "", "title": "" },
 ]
}
```

PR #1261, opened 2024-10-09
Digital Bookplates Upload App

### LINKED ISSUE #1187: Create UI for adding 979 fields
Staff will use a UI in airflow to upload a file of instance UUID's that they want to add 979 fields to. The upload form should present the user with a list of available bookplate fund metadata to add to records in FOLIO. This is what that form looks like in libsys-webforms:
Screenshot 2024-09-06 at 4 40 08 PM
The upload form will trigger the add 979 field DAG. We have a pattern for this interaction with the [boundwiths form](https://github.com/sul-dlss/libsys-airflow/blob/main/libsys_airflow/plugins/boundwith/boundwith_view.py). We should also keep track of the user who submitted the form so that they can be emailed when the add 979 field DAG is complete. The email should also go to sul-unicorn-devs list and sul_bulk_bookplates@lists.stanford.edu (confirm Tech Services still uses this list).

jermnelson:
Fixes #1187 

Current screenshots:
Screenshot 2024-10-09 at 8 19 15 AM
Screenshot 2024-10-09 at 8 19 35 AM

Blocked by #1186, stub in place for #1254.

### DIFF
--- a/tests/apps/digital_bookplates/test_batch_upload_view.py
+++ b/tests/apps/digital_bookplates/test_batch_upload_view.py
@@ -1,4 +1,7 @@
 import datetime
+import io
+
+import pandas as pd
 import pytest
 
 from airflow.www import app as application
@@ -10,6 +13,7 @@ from pytest_mock_resources import create_sqlite_fixture, Rows
 
 from libsys_airflow.plugins.digital_bookplates.apps.digital_bookplates_batch_upload_view import (
     DigitalBookplatesBatchUploadView,
+    _save_uploaded_file,
 )
 from libsys_airflow.plugins.digital_bookplates.models import DigitalBookplate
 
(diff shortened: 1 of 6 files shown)


tests/apps/digital_bookplates/test_batch_upload_view.py
```diff
@@ -0,0 +1,147 @@
...
+    ),
+    DigitalBookplate(
+        id=2,
+        created=datetime.datetime(2024, 10, 14, 17, 16, 15, 986798),
+        updated=datetime.datetime(2024, 10, 14, 17, 16, 15, 986798),
+        druid="ab123xy4567",
+        fund_name=None,
+        image_filename="ab123xy4567_00_0001.jp2",
+        title="Alfred E. Newman Magazine Fund for Humor Studies",
+        fund_uuid="f2421e3f-0089-497b-a3c6-104256a5c4ff",
```
shelleydoljack, line 40: There wouldn't be a fund_uuid if there is not a fund_name. Not sure if this will mess up the form and selection; it doesn't seem like it would.

shelleydoljack (approved):

tests/apps/digital_bookplates/test_batch_upload_view.py
jermnelson: Good point; I just pushed a commit that changes the test fixture so that a fund_uuid is set to None.

Issue #1260, opened 2024-10-08
Add some logging for new or updated bookplates


shelleydoljack:
https://github.com/sul-dlss/libsys-airflow/blob/d29dd5e2a3b0ff43174e9a540ff8e05b80b4e78e/libsys_airflow/plugins/digital_bookplates/purl_fetcher.py#L36-L57

When running this locally we won't get an email report with the new or updated bookplates. It would be nice to see in the log for each mapped task some info about the bookplate (new, update, no change).

For the extract_bookplate_metadata task, one of the mapped tasks we see:
![Screenshot 2024-10-08 at 3 32 29 PM](https://github.com/user-attachments/assets/0b914211-b34b-4549-9659-5d7f827c0468)
Should failure or fund_name show empty `{}` rather than, I'm guessing "None" in this context?

I think the add_update_model task is where we see in the XCOM the "newness" of a bookplate. This is from a first run of the DAG:
![Screenshot 2024-10-08 at 3 34 14 PM](https://github.com/user-attachments/assets/35b797a2-5214-42ad-9e4c-a5a8386fa78b)
And this is a subsequent run (so I guess `return_value = {}` means no change):
![Screenshot 2024-10-08 at 3 36 07 PM](https://github.com/user-attachments/assets/5fc4504e-aded-4b06-9d74-1ce4c6470f62)
Either way, it would be nice to have some `logger.info`.

shelleydoljack:
This came about because I was thinking I'd see the next DAG, digital_bookplate_instances, run but I didn't. I eventually saw in the log for `trigger_instances_dag` task the "No new funds to trigger digital_bookplate_instances DAG". 👍

PR #1267, opened 2024-10-09
Adds fund_uuid to new and updated lists.

### LINKED ISSUE #1265: Add fund_uuid to list of new funds that triggers the find instances DAG
The fund_uuid is added to the bookplates table with the _fetch_folio_fund_id function, which is called by [_add_bookplate and _update_bookplate](https://github.com/sul-dlss/libsys-airflow/blob/d29dd5e2a3b0ff43174e9a540ff8e05b80b4e78e/libsys_airflow/plugins/digital_bookplates/purl_fetcher.py#L49-L56). The issue is that in the context of the add_update_model task, the object passed to add or update the model, metadata, contains the Cocina data. At this point the fund_uuid is looked up in folio and written to the table, but metadata object doesn't contain the fund_uuid key and it is this object that is passed to the new and updates lists. See https://github.com/sul-dlss/libsys-airflow/blob/d29dd5e2a3b0ff43174e9a540ff8e05b80b4e78e/libsys_airflow/dags/digital_bookplates/fetch_digital_bookplates.py#L48-L49. When the next DAG [is triggered with the new list](https://github.com/sul-dlss/libsys-airflow/blob/d29dd5e2a3b0ff43174e9a540ff8e05b80b4e78e/libsys_airflow/plugins/digital_bookplates/purl_fetcher.py#L161-L169), the downstream task bookplate_funds_polines expects the funds list to contain the fund_uuid, see [_new_bookplates](https://github.com/sul-dlss/libsys-airflow/blob/d29dd5e2a3b0ff43174e9a540ff8e05b80b4e78e/libsys_airflow/plugins/digital_bookplates/bookplates.py#L41-L54).

shelleydoljack:
Fixes #1265

### DIFF
--- a/libsys_airflow/plugins/digital_bookplates/purl_fetcher.py
+++ b/libsys_airflow/plugins/digital_bookplates/purl_fetcher.py
@@ -47,12 +47,14 @@ def add_update_model(metadata) -> dict:
             .first()
         )
         if bookplate is None:
+            logger.info("Bookplate is new")
             (db_id, fund_uuid) = _add_bookplate(metadata, session)
             metadata["db_id"] = db_id
             metadata["fund_uuid"] = fund_uuid
             report["new"] = metadata
         else:
             metadata = _update_bookplate(metadata, bookplate, session)
+            logger.info("Bookplate was updated")
             if "reason" in metadata:
                 metadata["db_id"] = bookplate.id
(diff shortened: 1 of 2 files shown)


shelleydoljack:
Closes #1260

jgreben (approved):

Issue #1259, opened 2024-10-08
Debug invoice_lines_from_invoices task


shelleydoljack:
When running the digital_bookplate_instances DAG locally with okapi-dev as folio, it fails on the invoice_lines_from_invoices_task. The error in the log shows:
```
[2024-10-08, 18:33:33 UTC] {invoices.py:135} INFO - Getting invoice lines for 9a7749c6-2d4d-4d9f-9ed0-db121070f2b0
[2024-10-08, 18:33:33 UTC] {_client.py:1038} INFO - HTTP Request: GET https://okapi-dev.stanford.edu/invoice/invoice-lines?query=%28invoiceId%3D%3D9a7749c6-2d4d-4d9f-9ed0-db121070f2b0%29&limit=500&offset=0 "HTTP/1.1 200 OK"
[2024-10-08, 18:33:34 UTC] {_client.py:1038} INFO - HTTP Request: GET https://okapi-dev.stanford.edu/invoice/invoice-lines?query=%28invoiceId%3D%3D9a7749c6-2d4d-4d9f-9ed0-db121070f2b0%29&limit=500&offset=500 "HTTP/1.1 200 OK"
[2024-10-08, 18:33:34 UTC] {invoices.py:135} INFO - Getting invoice lines for 6c8bd41e-96e9-4321-a151-9402026d187a
[2024-10-08, 18:33:38 UTC] {taskinstance.py:2731} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.11/site-packages/httpx/_transports/default.py", line 72, in map_httpcore_exceptions
    yield
  File "/home/airflow/.local/lib/python3.11/site-packages/httpx/_transports/default.py", line 236, in handle_request
    resp = self._pool.handle_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.11/site-packages/httpcore/_sync/connection_pool.py", line 216, in handle_request
    raise exc from None
  File "/home/airflow/.local/lib/python3.11/site-packages/httpcore/_sync/connection_pool.py", line 196, in handle_request
    response = connection.handle_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.11/site-packages/httpcore/_sync/connection.py", line 101, in handle_request
    return self._connection.handle_request(request)
[...]

PR #1272, opened 2024-10-11
Adds FOLIOCLIENT_HTTP_TIMEOUT to increase default timeout to 10s.

shelleydoljack:
I'm not sure this fixes the issues identified in #1259. It might be a problem with using [FolioClient.folio_get_all](https://github.com/FOLIO-FSE/FolioClient/blob/034431fa586e787a487ab402446631e49ec701df/folioclient/FolioClient.py#L291) since when I try running the find bookplate instance DAG locally against test and stage with timeout 10s, the steps to get paid invoice ID's and then invoice lines IDs fails (almost immediately in the case of getting paid invoice IDs).


shelleydoljack:
I deployed this branch to airflow dev and ran the find bookplate instances dag. It errored with `httpx.RemoteProtocolError: Server disconnected without sending a response.` We could leave the timeout increased to 10s but I'm thinking we need to refactor how we are getting invoice IDs and invoice-line IDs.

jermnelson (commented):
Maybe bump this up to 30 seconds?

shelleydoljack:
I bumped to 30s and it still fails in the same way. https://sul-libsys-airflow-dev.stanford.edu/dags/digital_bookplate_instances/grid?root=&execution_date=2024-09-01T20%3A19%3A56%2B00%3A00&tab=logs&dag_run_id=manual__2024-09-01T20%3A19%3A56%2B00%3A00&task_id=invoice_lines_from_invoices

I'm going to try to address the issues with refactoring the task (ticket #1273 ). Maybe I open a bug issue in FolioClient if it turns out that the `folio_get_all` function is buggy.

Issue #1271, opened 2024-10-11
Don't lookup fund_uuid in folio when fund_name is None


shelleydoljack:
This is broken now. https://github.com/sul-dlss/libsys-airflow/blob/e6efd8041c4165763cc14ae57c38d959c28d1833/libsys_airflow/plugins/digital_bookplates/purl_fetcher.py#L230

When running the DAG, the log shows that for when fund_name is NONE, a request to folio is still made and not handled well.:
```
2024-10-11, 20:46:28 UTC] {_client.py:1038} INFO - HTTP Request: GET https://okapi-test.stanford.edu/finance/funds?query=name%3D%3DNone "HTTP/1.1 200 OK"
[2024-10-11, 20:46:28 UTC] {taskinstance.py:2731} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.11/site-packages/airflow/models/taskinstance.py", line 444, in _execute_task
    result = _execute_callable(context=context, **execute_callable_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.11/site-packages/airflow/models/taskinstance.py", line 414, in _execute_callable
    return execute_callable(context=context, **execute_callable_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.11/site-packages/airflow/decorators/base.py", line 241, in execute
    return_value = super().execute(context)
                   ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.11/site-packages/airflow/operators/python.py", line 200, in execute
    return_value = self.execute_callable()
                   ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.11/site-packages/airflow/operators/python.py", line 217, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/airflow/libsys_airflow/plugins/digital_bookplates/purl_fetcher.py", line 51, in add_update_model
[...]

PR #1275, opened 2024-10-14
Catches IndexError if Okapi call returns an empty list

jermnelson:
Discovered error when running `fetch_digital_bookplates` DAG locally. Okapi returns an empty list `[]` and trying to retrieve the first list element on line 237 was raising an `IndexError`.


shelleydoljack (approved):
This fixes #1271 . I see we'd need to handle empty lists from folio anyways when there is a fund_name but it is not in folio.

jgreben (approved):

PR #1279, opened 2024-10-15
Refactors invoice queries

### LINKED ISSUE #1273: Refactor tasks that get invoice ID's and invoice line ID's
See #1259 for some background.

It could be a problem with using the `folio_get_all` function. Some suggestions for refactoring: 
- use the storage endpoints instead
- expand the lists into mapped tasks (so that retries are not retrying the whole list again)
- try using folio_get instead of folio_get_all for the `_get_all_invoice_lines` (we realistically will probably not have more than 200 invoice lines per invoice)
- maybe we can keep folio_get_all for the `_get_all_ids_from_invoices` function?
- if we keep getting errors for the `_get_all_ids_from_invoices` function, then we should reconsider the DAG schedule when trying to run this with a from_date of 8/28/2023. Maybe we do `catchup=True,` in the DAG configuration in order to spawn many DAG runs with only weekly paymentDate ranges.

shelleydoljack:
Closes #1273
- uses the storage endpoints instead to reduce internal okapi calls
- expands the list of invoice UUIDs into mapped tasks and groups the filtering down of invoice lines to getting instances paid on bookplate funds
- `_get_all_invoice_lines` uses the FolioClient `folio_get_all` which ends up just using `folio_get`. The intermittent `httpx.RemoteProtocolError: Server disconnected without sending a response.` error still occurs but since the tasks are mapped into task groups, we have a more granular way to retry.

Screenshot 2024-10-15 at 10 57 20 AM

Payload to trigger add 979 DAG for each mapped task:
Screenshot 2024-10-15 at 10 58 19 AM

### DIFF
--- a/libsys_airflow/dags/digital_bookplates/digital_bookplate_instances.py
+++ b/libsys_airflow/dags/digital_bookplates/digital_bookplate_instances.py
@@ -1,6 +1,6 @@
 from datetime import datetime, timedelta
 
-from airflow.decorators import dag, task
+from airflow.decorators import dag, task, task_group
 from airflow.operators.empty import EmptyOperator
 from airflow.operators.python import get_current_context
 from airflow.timetables.interval import CronDataIntervalTimetable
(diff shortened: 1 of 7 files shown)


jermnelson (commented):
Looking at the `mypy` errors, I think there are a couple of options. The easiest would be add the ignore directive `# type: ignore` to lines 35 and 36 in the `digital_bookplate_instance` DAG. The second option would be to use a keyword argument for `invoice_lines` instead of a positional argument in the `bookplate_funds_polines` task and use kwargs for the `instances_from_po_lines` as well. I'm fine with either option.

jermnelson (approved):

PR #1303, opened 2024-10-17
Polls for status of 979 Fields DAG Runs and Emails Summary

### LINKED ISSUE #1269: Email report at the end a day from the polling DAG for successes and failures
Email body should contain a summary of 979 dag runs that were successes and failures. For the failures, we want a link to the dag run so we can go to airflow and investigate.

The summary for successful adding of 979 fields should include a total count of instances that had 979s added.

Failures: when a MARC record failed to update, include the instance ID and the 979 field (if accessible at this point in the dag).

When dag runs are triggered from the UI or the find instances DAG, send the dag run ID's to the polling DAG.

jermnelson:
Fixes #1269 

TODO:
- [x] Unit testing of emails
- [x] Test end-to-end on airflow-dev

### DIFF
--- a/libsys_airflow/plugins/digital_bookplates/apps/digital_bookplates_batch_upload_view.py
+++ b/libsys_airflow/plugins/digital_bookplates/apps/digital_bookplates_batch_upload_view.py
@@ -41,7 +41,7 @@ def _trigger_add_979_dags(**kwargs) -> str:
 
 
 def _trigger_poll_add_979_dags(dag_runs: list, email: str):
-    """Placeholder for polling DAGs DAG"""
+    """Triggers polling DAGs DAG"""
     dagbag = DagBag("/opt/airflow/dags")
     dag = dagbag.get_dag('poll_for_digital_bookplate_979s')
     execution_date = timezone.utcnow()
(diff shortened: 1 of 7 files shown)


libsys_airflow/plugins/digital_bookplates/apps/digital_bookplates_batch_upload_view.py
```diff
@@ -42,6 +42,18 @@ def _trigger_add_979_dags(**kwargs) -> str:
 
 def _trigger_poll_add_979_dags(dag_runs: list, email: str):
     """Placeholder for polling DAGs DAG"""
```
jgreben, line 44: Remove this comment?

jgreben (commented):
Non-blocking, looks like there is a comment that can be removed now that the function is no longer a placeholder?

libsys_airflow/plugins/digital_bookplates/apps/digital_bookplates_batch_upload_view.py
jermnelson: Thanks @jgreben, I updated the docstring to reflect what the function's purpose.

jgreben (approved):

PR #1317, opened 2024-10-22
[DRAFT] Multiple DAG triggers for Digital Bookplates Fetch and Instances DAGs

### LINKED ISSUE #1238: Trigger the add 979 fields DAG from digital_bookplate_instances DAG
At the end of the DAG for finding instances paid on bookplate funds, we should trigger the next DAG, which will add 979 data to MARC tags. This task is currently in a WIP branch: https://github.com/sul-dlss/libsys-airflow/blob/5a91e11a77b976a1ecd5e5cc5fe9e91c0f5326d1/libsys_airflow/dags/digital_bookplates/digital_bookplate_instances.py#L62

We have a ticket for starting the add 979 fields DAG here https://github.com/sul-dlss/libsys-airflow/issues/1186 that details a bit what the output of this task should be.

The XCOM from the previous task (in ticket https://github.com/sul-dlss/libsys-airflow/issues/1210) should have a data structure that looks like this (instance ID as key and one or more bookplate as value):
```
{
 "242c6000-8485-5fcd-9b5e-adb60788ca59": [
  	{ "druid": "", "fund_name": "", "image_filename": "", "title": "" },
	…
 ]
}
```
We should use this data to trigger the next DAG with fund names and instances.

jermnelson:
Fixes #1238 

Because the `fetch_digital_bookplates` DAG triggers one or more `digital_book_plates_instances` DAG runs, and
each `digital_book_plates_instances` DAG run in turn triggers a `digital_bookplate_979` DAG to add the 979 MARC fields, each `digital_bookplate_979`  DAG run id is generated and passed as a parameter to the `digital_book_plates_instances` DAG run. The generated `digital_bookplate_979`  DAG run ids are collected and then passed as a parameter to the triggered `poll_for_digital_bookplate_979s` in the `fetch_digital_bookplates`. 

The `poll_for_digital_bookplate_979s` polls for the status of all of the `digital_bookplate_979` DAG until they exist and the status is either success or failed. 

Here is a diagram of the DAG trigger interactions:

```mermaid
graph TD
fetch_digital_bookplates --> digital_bookplate_instances[1+ DAG runs]
digital_bookplate_instances[1+  digital_book_plates_instances DAG runs] --> digital_bookplate_979[1+ digital_bookplate_979]
fetch_digital_bookplates --> poll_for_digital_bookplate_979s
poll_for_digital_bookplate_979s --> digital_bookplate_979[1+ digital_bookplate_979]
```

TODO:
- [ ] Full end-to-end test 
- [ ] Unit testing of new code

### DIFF
--- a/libsys_airflow/dags/digital_bookplates/digital_bookplate_instances.py
+++ b/libsys_airflow/dags/digital_bookplates/digital_bookplate_instances.py
@@ -4,6 +4,7 @@ from airflow.decorators import dag, task_group
 from airflow.models.param import Param
 from airflow.operators.empty import EmptyOperator
 from airflow.timetables.interval import CronDataIntervalTimetable
+from airflow.utils import timezone
 
 
 from libsys_airflow.plugins.digital_bookplates.bookplates import (
@@ -50,7 +51,12 @@ def process_invoice_lines_group(invoice_id: str):
             format="datetime",
             type="string",
             description="The earliest date to select paid invoices from FOLIO.",
-        )
+        ),
+        "digital_bookplate_979_run_id": Param(
+            f"manual__{timezone.utcnow().isoformat()}",
+            type="string",
+            description="The run id for the digital_bookplate_979 DAG run",
+        ),
     },
 )
 def digital_bookplate_instances():
--- a/libsys_airflow/dags/digital_bookplates/fetch_digital_bookplates.py
+++ b/libsys_airflow/dags/digital_bookplates/fetch_digital_bookplates.py
@@ -18,6 +18,7 @@ from libsys_airflow.plugins.digital_bookplates.purl_fetcher import (
     fetch_druids,
     filter_updates_errors,
     trigger_instances_dag,
+    trigger_polling_979_dag,
 )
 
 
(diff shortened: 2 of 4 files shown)


shelleydoljack:
I'm not sure this is how this should work. I'm confused why fetch_digital_bookplates DAG would trigger the polling dag at all. In most cases, the digital_bookplate_instances DAG will run on its weekly schedule and not be triggered by fetch_digital_bookplates at all but in turn trigger the digital_bookplate_979 DAG. It seems in most cases, during the weekly scheduled run, the digital_bookplate_instances DAG would trigger the polling DAG to check for successes and failures in the digital_bookplate_979 DAGs. In the case where there are new bookplates we get from fetch_digital_bookplates, we will get an email already of the new bookplates. New bookplates will trigger digital_bookplate_instances DAG runs, which should then trigger digital_bookplate_979 DAG runs. So the polling DAG should get the run ID's to poll from digital_bookplate_instances, right?

jermnelson:
Closing because assumptions made about the DAG flows are incorrect.

PR #1344, opened 2024-10-24
Adds comments with function inputs and output direction

jgreben:
Also move the order of the functions in the file for more clarity.  When reviewing, it may be easier to look at the entire file rather than the diffs.

### DIFF
--- a/libsys_airflow/plugins/digital_bookplates/bookplates.py
+++ b/libsys_airflow/plugins/digital_bookplates/bookplates.py
@@ -133,7 +133,6 @@ def launch_add_979_fields_task(**kwargs):
 
 @task
 def add_979_marc_tags(druid_instances: dict) -> dict:
-    # Get the bookplate data from the bookplates table:
     """
     "242c6000-8485-5fcd-9b5e-adb60788ca59": [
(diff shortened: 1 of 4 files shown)


jermnelson (approved):

libsys_airflow/plugins/digital_bookplates/bookplates.py
```diff
@@ -97,49 +86,61 @@ def bookplate_funds_polines(**kwargs) -> list:
...
+    """
+    Trigger add a tag dag with instance UUIDs and fund 979 data. Returns a dict
+    """
+    params = kwargs.get("params", {})
+    return params.get("druids_for_instance_id", {})  # -> add_979_marc_tags
 
 
 @task
 def add_979_marc_tags(druid_instances: dict) -> dict:
+    # Get the bookplate data from the bookplates table:
```
shelleydoljack, line 136: This doesn't get the data from the table but rather the upstream tasks and dags get the data from the table. When I read this, I get the idea that this function gets the data from the table 🤷 .

shelleydoljack (approved):

PR #1364, opened 2024-10-25
Add logging for empty bookplates_polines

### LINKED ISSUE #1358: Add logging when no invoice lines were paid using bookplate funds for the week
In the weekly schedule run of digital_bookplate_instances, it could be that no bookplate funds were used. When this happens we currently see a return of `[]` in the log for bookplate_fund_polines and the tasks pass through the DAG successfully.
Screenshot 2024-10-25 at 11 25 22 AM

We should add some logging too. Maybe skip the downstream tasks or just let them finish successfully. 

https://github.com/sul-dlss/libsys-airflow/blob/610c930f7cd9bb284b90d061f123f2e154213314/libsys_airflow/plugins/digital_bookplates/bookplates.py#L130-L141

jgreben:
Fixes #1358

### DIFF
--- a/libsys_airflow/plugins/digital_bookplates/bookplates.py
+++ b/libsys_airflow/plugins/digital_bookplates/bookplates.py
@@ -138,6 +138,9 @@ def bookplate_funds_polines(**kwargs) -> list:
                         {"bookplate_metadata": bookplate, "poline_id": poline_id}
                     )
 
+    if len(bookplates_polines) == 0:
+        logger.info("No bookplate funds were used. Downstream tasks will be skipped.")
+    
     return bookplates_polines  # -> instances_from_po_lines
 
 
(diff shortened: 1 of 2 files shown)


libsys_airflow/plugins/digital_bookplates/bookplates.py
```diff
@@ -138,6 +138,9 @@ def bookplate_funds_polines(**kwargs) -> list:
                         {"bookplate_metadata": bookplate, "poline_id": poline_id}
                     )
 
+    if len(bookplates_polines) == 0:
+        logger.info("No bookplate funds were used. Downstream tasks will be skipped")
```
shelleydoljack, line 142: The downstream tasks don't actually get skipped, they complete successfully. We'd have to do something more to the dag file I think but it's not a big deal. I don't think we kick off dag runs if instances is empty.

jgreben: Ok I changed the message.

jermnelson (approved):

PR #1380, opened 2024-10-29
Skip OCLC records op if not prod

### LINKED ISSUE #1054: skip transmission for transmit_data_oclc_api_task when env is not production
Blocked by #988 

For the transmit_data_oclc_api_task, we're not calling the return_success_test_instance when the env is not production. 
https://github.com/sul-dlss/libsys-airflow/blob/8760131926c03da19ac19e6036819e8a550209f5/libsys_airflow/plugins/data_exports/transmission_tasks.py#L164-L183

We should add that once we have completed #988

jgreben:
Fixes #1054

### DIFF
--- a/libsys_airflow/plugins/data_exports/oclc_api.py
+++ b/libsys_airflow/plugins/data_exports/oclc_api.py
@@ -16,6 +16,7 @@ from libsys_airflow.plugins.data_exports.marc.oclc import get_record_id
 from libsys_airflow.plugins.data_exports.marc.excluded_tags import oclc_excluded
 
 from libsys_airflow.plugins.shared.folio_client import folio_client
+from libsys_airflow.plugins.shared.utils import is_production
 
 from pymarc import Record
 
@@ -36,7 +37,6 @@ def get_instance_uuid(record) -> Union[str, None]:
 
 
 def oclc_records_operation(**kwargs) -> dict:
-
     function_name: str = kwargs["oclc_function"]
     connection_lookup: dict = kwargs["connections"]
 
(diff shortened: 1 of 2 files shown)


jermnelson:
We may want to hold off merging this PR until we hear from @ahafele, I think she is still doing testing on Airflow test.

libsys_airflow/plugins/data_exports/oclc_api.py
```diff
@@ -44,30 +45,33 @@ def oclc_records_operation(**kwargs) -> dict:
...
-            failures[library].extend(oclc_result['failures'])
-            archive_files.extend(oclc_result['archive'])
-            logger.info(
-                f"Processed {function_name} for {library} successful {len(success[library])} failures {len(failures[library])}"
-            )
-        else:
-            logger.info(f"No {function_name} records for {library}")
+        logger.info(
+            f"Skipping OCLC API {function_name} record operation not in production"
+        )
```
shelleydoljack, line 74: I think we still need to put something in archive_files in order to move the selection files to the transmitted directory (even though we aren't transmitting them in dev and stage envs).

shelleydoljack (changes requested):

tests/data_exports/test_oclc_api.py
```diff
@@ -1028,21 +1033,37 @@ def test_oclc_records_operation_no_records(mock_oclc_api, caplog):
...
+    result = oclc_api.oclc_records_operation(
+        oclc_function="delete", connections=connections, records=test_records_dict
+    )
 
-    second_mod = oclc_api.__oclc_marc_modifications__(sample_records[1])
+    assert result['success'] == {'STF': []}
+    assert result['failures'] == {'STF': []}
 
-    assert second_mod["007"].value() == "cr un|||||a||||"
-    assert second_mod["040"]["a"] == "STF"
```
shelleydoljack, line 1042: These test are now missing. Is that right?

shelleydoljack (changes requested):
I think the test `test_oclc_marc_modifications` got dropped inadvertently.

tests/data_exports/test_oclc_api.py
jgreben: I rebased and then merged. The missing test is back.

shelleydoljack (approved):
👍

PR #1399, opened 2024-11-04
Add subject line for production vs other and refactor tests

jgreben:
Fixes https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/685
When sending fix_encubrances_script email, specify whether it is run in production or not.

### DIFF
--- a/libsys_airflow/plugins/folio/encumbrances/email.py
+++ b/libsys_airflow/plugins/folio/encumbrances/email.py
@@ -1,3 +1,5 @@
+import re
+
 from airflow.models import Variable
 from airflow.utils.email import send_email
 from jinja2 import Template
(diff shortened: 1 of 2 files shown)


libsys_airflow/plugins/folio/encumbrances/email.py
```diff
@@ -40,3 +42,11 @@ def _email_body(log):
     html_body = email_template.render(log_content=log.read())
 
     return html_body
+
+
+def subject(**kwargs):
+    library = kwargs.get("library", "")
+    if is_production():
+        return f"Fix Encumbrances for {library} - folio-prod"
```
shelleydoljack, line 50: In the other places where we are doing this, we put the env in front. But we also don't put the env if it is folio-prod and put the env if it is folio-test or stage. Could we do the same here? Here is an example of where we do this: https://github.com/sul-dlss/libsys-airflow/blob/9f481a6a61ef2c4fc582071cef6445be85b61d51/libsys_airflow/plugins/digital_bookplates/email.py#L241-L256

shelleydoljack (changes requested):

shelleydoljack (approved):
Thanks for making it consistent!

PR #1407, opened 2024-11-05
Use send_email wrapper to apply subject with server name when not prod

### LINKED ISSUE #1131: Include server name in any emails sent from airflow
When testing the feeder file for gsb a real email was generated and sent to the libraries. Including -stage -test -dev in Subject would help reduce false alarms.

jgreben:
Fixes #1131

### DIFF
--- a/tests/digital_bookplates/test_bookplates_emails.py
+++ b/tests/digital_bookplates/test_bookplates_emails.py
@@ -31,7 +31,7 @@ failures = [
 
 @pytest.fixture
 def mock_folio_variables(monkeypatch):
-    def mock_get(key):
+    def mock_get(key, *args):
         value = None
         match key:
             case "FOLIO_URL":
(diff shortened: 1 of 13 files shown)


tests/digital_bookplates/test_bookplates_emails.py
jgreben, line 1: @jermnelson Maybe you can help figure out why the test fail for this test file?

jgreben: When patching `"libsys_airflow.plugins.digital_bookplates.email.send_email_with_server_name"` it never calls the wrapper function when running the test. If you change the patch to `"libsys_airflow.plugins.shared.utils.send_email_with_server_name"` it calls the wrapper function, but then it tries to send a real email and does not actually patch `send_email`.

jermnelson (approved):

shelleydoljack (approved):

PR #1422, opened 2024-11-08
Refactor to use dag_run_url Function in utils

### LINKED ISSUE #1417: DAG run URL incorrect
The link https://sul-libsys-airflow-prod.stanford.edu/dags/86302/grid?dag_run_id=scheduled__2024-11-06T08%3A00%3A00%2B00%3A00 goes to error message about 86302 not in the DagBag. 

I think `dag_run.id` is not the same as `dag_run.dag_id`. dag_id for this should be send_pod_records.

https://github.com/sul-dlss/libsys-airflow/blob/dd9f90aa058244a8c696d25a8739d92b9297f0e3/libsys_airflow/plugins/data_exports/email.py#L95

jermnelson:
Fixes #1417 

Dag run urls in digital_bookplates and data_export plugins now use a single utils function, `dag_run_url`.

### DIFF
--- a/libsys_airflow/plugins/data_exports/email.py
+++ b/libsys_airflow/plugins/data_exports/email.py
@@ -1,6 +1,5 @@
 import logging
 import pathlib
-import urllib
 
 from jinja2 import Template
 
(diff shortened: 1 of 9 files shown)


shelleydoljack (approved):
Thanks for DRY'ing it up!

PR #1423, opened 2024-11-08
Remove scheduling, catchup, and start-date from OCLC DAGs

### LINKED ISSUE #1419: OCLC - remove schedule during initial catch up period
Remove schedule during catch up period and then re-add when ready to run daily.

jermnelson:
Fixes #1419. Removed catchup and start-date because we will be doing manual DAG runs to catchup for past year in FOLIO and avoid exceeding OCLC API usage caps.

### DIFF
--- a/libsys_airflow/dags/data_exports/oclc_selections.py
+++ b/libsys_airflow/dags/data_exports/oclc_selections.py
@@ -7,9 +7,7 @@ from airflow.decorators import task, task_group
 from airflow.models.param import Param
 from airflow.operators.python import BranchPythonOperator
 from airflow.operators.empty import EmptyOperator
-from airflow.models import Variable
 from airflow.operators.python import PythonOperator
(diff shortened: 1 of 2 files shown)


shelleydoljack (changes requested):
I'm almost certain you need to pass `schedule=None` or else it defaults to everyday at midnight UTC (at least that is what I noticed was happening with https://github.com/sul-dlss/libsys-airflow/commit/25624de1413aee1d2cfcab494493230839a15ea3)

jermnelson:
> I'm almost certain you need to pass `schedule=None` or else it defaults to everyday at midnight UTC (at least that is what I noticed was happening with [25624de](https://github.com/sul-dlss/libsys-airflow/commit/25624de1413aee1d2cfcab494493230839a15ea3))

Done in the latest commit.

shelleydoljack (approved):

PR #1426, opened 2024-11-11
Log errors but do not raise SystemExit for encumbrances script

jgreben:
The fix_encumbrances dag will not email the log if there is an error that raises SystemExit(1). Log the errors and always complete the script.

### DIFF
--- a/libsys_airflow/plugins/folio/encumbrances/fix_encumbrances.py
+++ b/libsys_airflow/plugins/folio/encumbrances/fix_encumbrances.py
@@ -48,7 +48,7 @@ def login(tenant, username, password):
         r = requests.post(okapi_url + '/authn/login', headers=login_headers, json=data)
         if r.status_code != 201:
             raise_exception_for_reply(r)
-        print('Logged in successfully.')
+        logger.info('Logged in successfully.')
         okapi_token = r.json()['okapiToken']
         return {
             'x-okapi-tenant': tenant,
--- a/libsys_airflow/plugins/folio/encumbrances/fix_encumbrances.py
+++ b/libsys_airflow/plugins/folio/encumbrances/fix_encumbrances.py
@@ -48,7 +48,7 @@ def login(tenant, username, password):
         r = requests.post(okapi_url + '/authn/login', headers=login_headers, json=data)
         if r.status_code != 201:
             raise_exception_for_reply(r)
-        logger.info('Logged in successfully.')
+        print('Logged in successfully.')
         okapi_token = r.json()['okapiToken']
         return {
             'x-okapi-tenant': tenant,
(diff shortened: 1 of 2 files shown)


libsys_airflow/plugins/folio/encumbrances/fix_encumbrances.py
```diff
@@ -56,8 +56,8 @@ def login(tenant, username, password):
             'Content-Type': 'application/json',
         }
     except Exception as err:
-        logger.info('Error during login:', err)
-        raise SystemExit(1)
+        logger.error('Error during login:', err)
+        # raise SystemExit(1)
```
jermnelson, line 60: Any reason why to keep this line commented out, `# SystemExit(1)` instead of just removing here and the other similar lines in this PR?

libsys_airflow/plugins/folio/encumbrances/fix_encumbrances.py
```diff
@@ -67,11 +67,11 @@ async def get_request_without_query(url: str) -> dict:
         if resp.status_code == HTTPStatus.OK:
             return resp.json()
         else:
-            logger.info(f'Error getting record with url {url} : \n{resp.text} ')
-            raise SystemExit(1)
+            logger.error(f'Error getting record with url {url} : \n{resp.text} ')
+            # raise SystemExit(1)
```
jermnelson, line 71: See comment above

jermnelson (changes requested):

libsys_airflow/plugins/folio/encumbrances/fix_encumbrances.py
jgreben: Because we will likely need to revisit and re-implement this script when there is a new version of it, I would like to keep a documentation of the edits to make it easier to re-do what we did to make it work.

libsys_airflow/plugins/folio/encumbrances/fix_encumbrances.py
jgreben: Same answer as above

jermnelson (approved):

jgreben:
I modified the fix_encumbrances script to print the "info" and "warning" statements to stdout (the log file) and the "errors" to both stdout and the logger.

libsys_airflow/plugins/folio/encumbrances/fix_encumbrances.py
```diff
@@ -723,9 +750,9 @@ async def unrelease_encumbrances_with_non_zero_amounts(
 async def unrelease_open_orders_encumbrances_with_nonzero_amounts(
     fiscal_year_id, open_orders_ids
 ):
-    logger.info('Unreleasing open orders encumbrances with non-zero amounts...')
+    print('Unreleasing open orders encumbrances with non-zero amounts...')
```
shelleydoljack, line 753: Is this the one where we updated it to fix it even if it has zero amounts? I think we should fixup the warning message too.

shelleydoljack (approved):
Do you want to address the non-zero amounts thing here or in a different PR? I'm okay either way.

libsys_airflow/plugins/folio/encumbrances/fix_encumbrances.py
jgreben: I think that was https://github.com/sul-dlss/libsys-airflow/blob/0bfa734681cf92ff820197f6f185bb48b289cb3c/libsys_airflow/plugins/folio/encumbrances/fix_encumbrances.py#L491 ?

Issue #1451, opened 2024-11-21
MARC files not getting state of transmitted after send DAG finishes


ahafele:
I am seeing this with OCLC after a successful run today - https://sul-libsys-airflow-prod.stanford.edu/data_export_download/. I believe if I try to run OCLC again, these files will be sent a second time. 

But I think it might also be the case for GOBI?

shelleydoljack:
I logged into the prod server and took a look at gobi. I think there is something more going on. It looks like the new files are being moved to transmitted but the updates are not. ANd looking at the logs for the last 2 scheduled runs of send_gobi, it looks like only new files are being sent.
```
libsys@sul-libsys-airflow-prod:~/libsys-airflow/current/data-export-files/gobi$ ls -l instanceids/updates/
total 17344
-rw-rw-r-- 1 libsys root 4351496 Aug 27 18:14 202408280114.csv
-rw-rw-r-- 1 libsys root  817663 Aug 29 08:21 202408291521.csv
-rw-rw-r-- 1 libsys root  929773 Sep  3 22:31 202409040531.csv
-rw-rw-r-- 1 libsys root  811817 Sep 10 22:31 202409110531.csv
-rw-rw-r-- 1 libsys root  779997 Sep 17 22:31 202409180531.csv
-rw-rw-r-- 1 libsys root 1054426 Sep 24 22:31 202409250531.csv
-rw-rw-r-- 1 libsys root  898249 Oct  1 22:31 202410020531.csv
-rw-rw-r-- 1 libsys root 1046693 Oct  8 22:32 202410090532.csv
-rw-rw-r-- 1 libsys root  674880 Oct 15 22:31 202410160531.csv
-rw-rw-r-- 1 libsys root  977133 Oct 22 22:31 202410230531.csv
-rw-rw-r-- 1 libsys root  720464 Oct 29 22:31 202410300531.csv
-rw-rw-r-- 1 libsys root 2697855 Nov  5 22:34 202411060634.csv
-rw-rw-r-- 1 libsys root  850704 Nov 12 22:30 202411130630.csv
-rw-rw-r-- 1 libsys root 1058089 Nov 19 22:31 202411200631.csv
libsys@sul-libsys-airflow-prod:~/libsys-airflow/current/data-export-files/gobi$ ls -l instanceids/new
total 0
libsys@sul-libsys-airflow-prod:~/libsys-airflow/current/data-export-files/gobi$ ls -l marc-files/new/
total 75420
[...]

PR #1454, opened 2024-11-21
Refactor OCLC Transmissions to Archive all Files

jermnelson:
Partial fix for #1451

### DIFF
--- a/libsys_airflow/plugins/data_exports/oclc_api.py
+++ b/libsys_airflow/plugins/data_exports/oclc_api.py
@@ -55,27 +55,23 @@ def oclc_records_operation(**kwargs) -> dict:
 
             success[library] = []
             failures[library] = []
-            archive_files = []
             if len(records) > 0:
                 oclc_result = oclc_api_function(records)
                 success[library].extend(oclc_result['success'])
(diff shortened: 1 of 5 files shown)


shelleydoljack (approved):

libsys_airflow/plugins/data_exports/oclc_api.py
```diff
@@ -55,27 +55,23 @@ def oclc_records_operation(**kwargs) -> dict:
...
-        archive_files = []
         for library, records in type_of_records.items():
             for record in records:
-                archive_files.append(record)
                 logger.info(
                     f"Skipping OCLC API {function_name} record operation not in production"
                 )
 
-    return {"success": success, "failures": failures, "archive": archive_files}
+    return {"success": success, "failures": failures}
```
shelleydoljack, line 74: Doesn't this mean for the no-op we need to put all the files in success?

jermnelson: I don't think so; since this PR is moving all gathered files regardless of status to transmitted sub-directories, the no-op doesn't impact the downstream archive tasks.

Issue #1445, opened 2024-11-19
use full dump selection for backstage initial export


shelleydoljack:
Things we want configurable in the full dump selection:
- use or not use junk tags
- exclude or select library (or campus?)
- cataloged date is configurable (we want only since FOLIO migration) (this is probably already like this in the full dump)

Selection Criteria:
Initial

- All libraries, excluding Lane
- New Bib records
- Instance status term = Cataloged
- Cataloged date = real date (not empty)
- Unsuppressed at Instance level
- Using specified timeframe, e.g. since FOLIO migration
- No deletes
- No updates

Record processing:
- do not do the "junk tags" (keep the junk tags)

Exclude records with
915 |aNO EXPORT|bAUTHORITY VENDOR
or
590 |a MARCit brief record.

shelleydoljack:
Sample 999 in current full dump:
`=999  \\$aDR593 .U48 2020$hBook$lSAL3-STACKS$wLibrary of Congress classification$j1`
subfield l contains library

shelleydoljack:
Per meeting, we don't have to exclude Lane for the initial export.

ahafele:
Reviewed today and @jgreben says idea still sound.
> Could add a date parameter to full dump dag and create a different materialized view and parameterize the name in airflow so we can choose which one to run.

jgreben:
@ahafele  @shelleydoljack  Did we decide if we wanted the ability to exclude or select library, or campus, or both?

ahafele:
We decided that for backstage we don't need to exclude a library. However, there may be a future need to exclude a library from a full dump, e.g. if Lane decides to not participate in the cc0 export.

jgreben:
@ahafele Can I make a separate ticket for excluding a library and put it in the backlog for now?

dlrueda:
I'd say yes to making a separate ticket.  But it might not be in the backlog for long. Tom initially wanted the CC0 full dump to be done winter quarter.  I told him if we had to exclude a library it wouldn't be worked on til after we get prod to Quesnelia.

ahafele:
Per standup today, we will need to exclude records by campus sooner rather than later.

shelleydoljack:
[sad]   Shelley Doljack reacted to your message:
________________________________

jgreben:
[...]

PR #1505, opened 2025-02-12
Catalog date is configurable

jgreben:
Partially completes https://github.com/sul-dlss/libsys-airflow/issues/1445 by allowing the configuration of the Cataloged date and specifying whether we want to recreate the materialized view when the DAG is run.


shelleydoljack (commented):
I think we need to still be able to create the materialized view with catalogedDate is not null. I don't see that is possible with this PR.

jgreben:
> I think we need to still be able to create the materialized view with catalogedDate is not null. I don't see that is possible with this PR.

@shelleydoljack If it's ok with you, let me make a separate PR for that.

jgreben:
> I think we need to still be able to create the materialized view with catalogedDate is not null. I don't see that is possible with this PR.

@shelleydoljack I changed it so that the `from_date` is always at least "2023-08-23", so the sql will always at least be `'catalogedDate' between '2023-08-23' and <today>` and thus will never be null.

shelleydoljack (approved):
👍

jermnelson (approved):

PR #1507, opened 2025-02-13
Ignore or Fixes flake8 Linting Errors

jermnelson:
- Ignoring all **D** related linting errors related to missing `doctstrings` at the module, class, and function/method levels. 
- Ignoring all **I** [isort](https://pycqa.github.io/isort/) linting errors
- Ignore all S101 errors as `assert` is used extensively in our tests
- Individual fixes/ignore remaining **S** (Security)
- Remaining linting errors either ignored or fixed.

### DIFF
--- a/libsys_airflow/plugins/data_exports/apps/data_export_upload_view.py
+++ b/libsys_airflow/plugins/data_exports/apps/data_export_upload_view.py
@@ -93,7 +93,7 @@ class DataExportUploadView(AppBuilderBaseView):
             except Exception as e:
                 flash(f"Error: {e}")
             finally:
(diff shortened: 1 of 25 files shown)


jgreben (changes requested):
Can we just add `libsys_airflow/plugins/folio/encumbrances/fix_encumbrances*.py` to the flake8 exclude list in `setup.cfg`? If it does not accept a wildcard then just exclude `libsys_airflow/plugins/folio/encumbrances/fix_encumbrances.py` for now? (There is a draft PR to add a `fix_encumbrances_quesnelia.py` file.)

jgreben (approved):

PR #1522, opened 2025-03-07
Create full record dump for specified campuses

jgreben:
- Adds an additional indexed materialized view to use as a right-join on the full record dump materialized view in order to filter the result set to a specific set of campus codes.
- Include as a runtime parameter a string of campus codes to include, with a default set (all of them)
- Use a default from_date
- Update the discoverySuppressed condition to account for NULL table values.


libsys_airflow/plugins/data_exports/sql/filter_campus_ids.sql
```diff
@@ -0,0 +1,12 @@
...
+create materialized view filter_campus_ids as
+select distinct(instanceid) from sul_mod_inventory_storage.holdings_record
+where  permanentlocationid in (
+    select id from sul_mod_inventory_storage.location
+    where campusid in (
+        select id from sul_mod_inventory_storage.loccampus 
+        where jsonb->>'code' in (%(campuses)s)
+    )
+);
+CREATE UNIQUE INDEX filter_full_dump_campus_idx ON filter_campus_ids (instanceid);
```
shelleydoljack, line 12: Does this need to be dropped before create like the materialized view is dropped?

jermnelson (approved):

libsys_airflow/plugins/data_exports/sql/materialized_view.sql
```diff
@@ -15,9 +15,13 @@ and (I.jsonb->>'statusId')::uuid in (
...
 join sul_mod_source_record_storage.marc_records_lb M
   on M.id = R.id
+right join filter_instance_ids F
+  on F.instanceid = I.id
 order by I.jsonb->'hrid'
-;
\ No newline at end of file
+;
+CREATE UNIQUE INDEX data_export_marc_ids ON data_export_marc (id);
+CREATE UNIQUE INDEX data_export_marc_hrids ON data_export_marc (hrid);
```
shelleydoljack, line 27: Maybe these need to be dropped too? Or maybe do `REINDEX INDEX data_export_marc_ids`?

shelleydoljack (commented):
Some suggestions about the indexes. Overall LGTM.

PR #1539, opened 2025-03-31
Full dump sql binding with psycopg2

jgreben:
- Adds param for s3 bucket path (Marc-files or CC0)
- Use psycopg2 connection to create materialized view of campus ids in order to correctly bind params
- Adjust sql and subsequent queries for materialized views
- remove unnecessary test class

### DIFF
--- a/libsys_airflow/dags/data_exports/full_dump_retrieval.py
+++ b/libsys_airflow/dags/data_exports/full_dump_retrieval.py
@@ -84,6 +84,11 @@ with DAG(
             type="string",
             description="Comma-seperated list of campus coded to include in full dump selection.",
         ),
+        "bucket": Param(
+            "marc-files",
+            type="string",
+            description="The S3 bucket to deposit the MARC records. CC0 or marc-files.",
+        ),
     },
 ) as dag:
 
--- a/libsys_airflow/dags/data_exports/full_dump_retrieval.py
+++ b/libsys_airflow/dags/data_exports/full_dump_retrieval.py
@@ -84,10 +84,10 @@ with DAG(
             type="string",
             description="Comma-seperated list of campus coded to include in full dump selection.",
         ),
-        "bucket": Param(
+        "marc_file_dir": Param(
             "marc-files",
             type="string",
-            description="The S3 bucket to deposit the MARC records. CC0 or marc-files.",
+            description="The S3 marc file path to deposit the MARC records. CC0 or marc-files.",
             enum=["marc-files", "CC0"],
         ),
     },
--- a/libsys_airflow/dags/data_exports/full_dump_retrieval.py
+++ b/libsys_airflow/dags/data_exports/full_dump_retrieval.py
@@ -63,7 +63,7 @@ with DAG(
             description="Remove excluded tags listed in marc/excluded_tags.pyfrom incoming record.",
         ),
         "from_date": Param(
-            Variable.get("EPOCH_DATE", "1885-11-11"),
+            Variable.get("FOLIO_EPOCH_DATE", "1885-11-11"),
             format="date",
             type="string",
             description="The earliest date to select record IDs from FOLIO.",
--- a/libsys_airflow/dags/data_exports/full_dump_retrieval.py
+++ b/libsys_airflow/dags/data_exports/full_dump_retrieval.py
@@ -88,6 +88,7 @@ with DAG(
             "marc-files",
             type="string",
             description="The S3 bucket to deposit the MARC records. CC0 or marc-files.",
+            enum=["marc-files", "CC0"],
         ),
     },
 ) as dag:
--- a/libsys_airflow/dags/data_exports/full_dump_retrieval.py
+++ b/libsys_airflow/dags/data_exports/full_dump_retrieval.py
@@ -63,7 +63,7 @@ with DAG(
             description="Remove excluded tags listed in marc/excluded_tags.pyfrom incoming record.",
         ),
         "from_date": Param(
-            Variable.get("FOLIO_EPOCH_DATE", "2023-08-23"),
+            Variable.get("EPOCH_DATE", "1885-11-11"),
             format="date",
             type="string",
             description="The earliest date to select record IDs from FOLIO.",
--- a/libsys_airflow/plugins/data_exports/full_dump_marc.py
+++ b/libsys_airflow/plugins/data_exports/full_dump_marc.py
@@ -134,11 +134,11 @@ def reset_s3(**kwargs) -> None:
     context = get_current_context()
     params = context.get("params", {})  # type: ignore
     reset = params.get("reset_s3", True)
-    bucket = params.get("bucket", "marc-files")
+    marc_file_dir = params.get("marc_file_dir", "marc-files")
 
     if reset:
         bucket = Variable.get("FOLIO_AWS_BUCKET", "folio-data-export-prod")
-        s3_dir = s3_bucket_path(bucket)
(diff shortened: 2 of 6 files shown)


libsys_airflow/dags/data_exports/full_dump_retrieval.py
```diff
@@ -84,6 +84,11 @@
             type="string",
             description="Comma-seperated list of campus coded to include in full dump selection.",
         ),
+        "bucket": Param(
+            "marc-files",
+            type="string",
+            description="The S3 bucket to deposit the MARC records. CC0 or marc-files.",
+        ),
```
shelleydoljack, line 91: Can this be a drop-down list instead so we know the value to pass? It wasn't clear to me that "CC0 or marc-files" were the values until I looked at the plugin code.

shelleydoljack: There's an example of this in the [full dump transmission](https://github.com/sul-dlss/libsys-airflow/blob/2611c91c0d13a015c2f4cec67912bcbce1201c54/libsys_airflow/dags/data_exports/full_dump_transmission.py#L65-L71) dag.

libsys_airflow/plugins/data_exports/full_dump_marc.py
```diff
@@ -172,3 +153,10 @@ def add_quotes(csv_string):
     values = csv_string.split(',')
     quoted_values = ["'{}'".format(v.strip()) for v in values]
     return ','.join(quoted_values)
```
shelleydoljack, line 155: Can you please add what type is returned from this function?

libsys_airflow/plugins/data_exports/full_dump_marc.py
```diff
@@ -117,7 +97,7 @@ def fetch_full_dump_marc(**kwargs) -> str:
     batch_size = kwargs.get("batch_size", 1000)
     connection = kwargs.get("connection")
     cursor = connection.cursor()  # type: ignore
-    sql = "SELECT id, hrid, content FROM public.data_export_marc ORDER BY hrid LIMIT (%s) OFFSET (%s)"
+    sql = "SELECT instanceid, hrid, content FROM public.data_export_marc ORDER BY hrid LIMIT (%s) OFFSET (%s)"
```
shelleydoljack, line 100: 'Tis good to be specific! Thanks!

tests/data_exports/test_full_dump_selections.py
```diff
@@ -34,19 +48,6 @@ def cursor(self):
...
-    connection: MockConnection = MockConnection()
-
-    def pool(self):
-        return self
-
-    def getconn(self):
-        return self.connection
-
-    def putconn(self, connection):
-        return None
```
shelleydoljack, line 47: Does this mean we aren't using a connection pool anymore to get the records out?

shelleydoljack (changes requested):

tests/data_exports/test_full_dump_selections.py
jgreben: No it just means that we do not have to mock it for the tests. https://github.com/sul-dlss/libsys-airflow/blob/334a371c8eb2c28056765d354b2c9b3e6939b7e6/libsys_airflow/dags/data_exports/full_dump_retrieval.py#L198

libsys_airflow/dags/data_exports/full_dump_retrieval.py
```diff
@@ -63,7 +63,7 @@
             description="Remove excluded tags listed in marc/excluded_tags.pyfrom incoming record.",
         ),
         "from_date": Param(
-            Variable.get("FOLIO_EPOCH_DATE", "2023-08-23"),
+            Variable.get("EPOCH_DATE", "1885-11-11"),
```
shelleydoljack, line 66: Ahhh, now I remember we did have a param for "since FOLIO migration" so that we can do the Backstage full dump. Changing this here should impact that selection.

shelleydoljack: So when we want a CC0 dump or POD full dump, we will use the default EPOCH_DATE but when we want a Backstage full dump, we need to remember the migration date. I guess that's okay since we probably won't be doing annual Backstage full dumps. 🤷

libsys_airflow/plugins/data_exports/marc/exporter.py
```diff
@@ -183,8 +184,11 @@ def write_marc(
         """
         Writes marc record to a file system (local or S3)
         """
+        context = get_current_context()
+        params = context.get("params", {})  # type: ignore
+        bucket_path = params.get("bucket", "marc-files")
         marc_file_name = instance_file.stem
-        directory = marc_directory / "marc-files"
+        directory = marc_directory / bucket_path
```
shelleydoljack, line 191: This is confusing. In the context of the full dump, bucket_path will be "marc-files" in S3 or "CC0" in S3, but for our daily exports, it should be "marc-files" in local storage. So how does this know if it "marc-files" in S3 or local?

shelleydoljack (changes requested):
I'm not sure about the function that writes to S3 or local storage.

libsys_airflow/dags/data_exports/full_dump_retrieval.py
jgreben: It's either that or have two different SQL files and rework everything to choose the file with the appropriate statement.

libsys_airflow/plugins/data_exports/marc/exporter.py
jgreben: Are you confused about the variable name or something else? It defaults to 'marc-files' for the local storage case because there will not be any "bucket" param in those dag runs.

libsys_airflow/dags/data_exports/full_dump_retrieval.py
jgreben: We will be doing quarterly CC0 dumps so I thought the "Stanford founding date" would be a better default. Maybe it would be better to keep the Variable name as `FOLIO_EPOCH_DATE` though.

libsys_airflow/plugins/data_exports/marc/exporter.py
jgreben: Would it be clearer if I changed `bucket_path` to something else?

jgreben: I changed it to `marc_file_dir` and then I was able to remove the `s3_bucket_path` function in libsys_airflow/plugins/data_exports/full_dump_marc.py

shelleydoljack: This is clearer now. Thanks!

shelleydoljack:
In this case statement https://github.com/sul-dlss/libsys-airflow/blob/7b8aff7620fccfe93b69911498d2f35416b5f744/libsys_airflow/plugins/data_exports/marc/exporter.py#L77-L91 we want the backstage full-dump to do the check_915_authority function, but it looks like it will do the check_915 function instead.

jgreben:
> In this case statement
> 
> https://github.com/sul-dlss/libsys-airflow/blob/7b8aff7620fccfe93b69911498d2f35416b5f744/libsys_airflow/plugins/data_exports/marc/exporter.py#L77-L91
> 
> we want the backstage full-dump to do the check_915_authority function, but it looks like it will do the check_915 function instead.

I think we decided that we didn't need to work on this for a one-off backstage run.

shelleydoljack (approved):
🎉

PR #1549, opened 2025-04-30
Ensure that Variable is an int

jermnelson:
When `MAX_ENTITIES` is set in the UI it is coming through as a string instead of a int causing the batch function to not batch correctly.


shelleydoljack (approved):
I was reading through the [source for Variable](https://airflow.apache.org/docs/apache-airflow/2.8.3/_modules/airflow/models/variable.html#Variable). It looks like it will always be a string so maybe just cast to int instead of checking if it's not an int. 🤷

jermnelson:
> I was reading through the [source for Variable](https://airflow.apache.org/docs/apache-airflow/2.8.3/_modules/airflow/models/variable.html#Variable). It looks like it will always be a string so maybe just cast to int instead of checking if it's not an int. 🤷

Done, just pushed up a commit that always casts `MAX_ENTITIES` to an int.

PR #1568, opened 2025-06-19
Add columns to sal3_items.sql

jgreben:
Fixes https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/727


libsys_airflow/plugins/data_exports/sal3_items.py
```diff
@@ -41,7 +41,20 @@ def sal3_items_sql_file(**kwargs) -> Path:
...
+    (
+        (
+            item_uuid,
+            holding_id,
+            item_barcode,
+            item_status,
+            item_permanent_location,
+            item_temporary_location,
+            holdings_permanent_location,
+            item_temporary_location,
```
dlrueda, line 53: hmm, now item_temporary_location is here twice.  Once is enough, but now I'm confused why Jeanette is asking for it if it was already there...

jgreben: Hmmm, yeah. I removed the duplicate.

dlrueda (approved):

PR #1571, opened 2025-06-24
Retry add marc tags if first put does not work

### LINKED ISSUE #1563: Retry adding a tag if tag not present in subsequent GET to SRS
https://github.com/sul-dlss/libsys-airflow/blob/6d75f151e6fde4c356545b38a87ca80ac43c5a73/libsys_airflow/plugins/shared/utils.py#L102-L129

Here we get 202 accepted when the mod-source-record-manager module returns a response that it accepted our request to PUT the record. But we don't actually know whether the record got updated or not at this point. The module sends a message through Kafka to mod-srs to update the record. Sometimes mod-source-record-storage times out and in this case for dag run https://sul-libsys-airflow-stage.stanford.edu/dags/digital_bookplate_979/grid?dag_run_id=manual__2025-06-11T09%3A00%3A18.859926%2B00%3A00&tab=logs&task_id=add_marc_tags_to_record it did timeout and not actually update the [record in folio-stage](https://folio-stage.stanford.edu/inventory/view/5acaa044-f087-4609-9309-1823c99a761a?filters=staffSuppress.false). The log in mod-srm shows:
```
"2025-06-11 09:00:28.053 [vert.x-eventloop-thread-1] [171980/change-manager] [sul] [07e8be0e-1cf5-5061-b5b3-4517d9a74d77] [mod_source_record_manager] DEBUG ngParametersProvider initializeParameters:: initializing mapping parameters..."
"2025-06-11 09:00:28.634 [vert.x-eventloop-thread-1] [171980/change-manager] [sul] [07e8be0e-1cf5-5061-b5b3-4517d9a74d77] [mod_source_record_manager] INFO  tProducerServiceImpl sendEventInternal:: Event with type QM_RECORD_UPDATED was sent to kafka"
"2025-06-11 09:16:06.840 [vert.x-worker-thread-10] [171980/change-manager] [sul] [07e8be0e-1cf5-5061-b5b3-4517d9a74d77] [] DEBUG KafkaConsumerWrapper businessHandlerCompletionHandler:: Consumer - id: 5 subscriptionPattern: SubscriptionDefinition(eventType=QM_ERROR, subscriptionPattern=cluster\\.Default\\.\\w{1,}\\.QM_ERROR) Starting business completion handler, globalLoadSensor: 1"
"2025-06-11 09:16:06.854 [vert.x-worker-thread-2] [171980/change-manager] [sul] [07e8be0e-1cf5-5061-b5b3-4517d9a74d77] [] INFO  tProducerServiceImpl sendEventInternal:: Event with type QM_COMPLETED was sent to kafka"
"2025-06-11 09:16:07.061 [vert.x-worker-thread-17] [171980/change-manager] [sul] [07e8be0e-1cf5-5061-b5b3-4517d9a74d77] [] DEBUG KafkaConsumerWrapper businessHandlerCompletionHandler:: Consumer - id: 5 subscriptionPattern: SubscriptionDefinition(eventType=QM_ERROR, subscriptionPattern=cluster\\.Default\\.\\w{1,}\\.QM_ERROR) Committing offset: 1"
```
And the related requst in mod-srs log shows:
```
"2025-06-11 09:16:06.761 [vert.x-worker-thread-1] [171980/change-manager] [sul] [07e8be0e-1cf5-5061-b5b3-4517d9a74d77] [] WARN  uickMarcKafkaHandler handle:: Failed to handle QM_RECORD_UPDATED event"
"2025-06-11 09:16:06.819 [vert.x-worker-thread-16] [171980/change-manager] [sul] [07e8be0e-1cf5-5061-b5b3-4517d9a74d77] [] INFO  uickMarcKafkaHandler sendEventToKafka:: Event with type QM_ERROR was sent to kafka"
```

jgreben:
Fixes #1563

### DIFF
--- a/libsys_airflow/plugins/shared/utils.py
+++ b/libsys_airflow/plugins/shared/utils.py
@@ -134,12 +134,16 @@ class FolioAddMarcTags(object):
             ).json()
             srs_fields = srs_update["records"][0]["parsedRecord"]["content"]["fields"]
             if self.__check_retry_put__(srs_fields, marc_instance_tags):
-                if self.retry_count < 1:
+                if self.retry_count < int(
+                    Variable.get("ADD_MARC_TAG_RETRY_COUNT", "1")
+                ):
                     self.retry_count += 1
                     logger.info(
                         f"Making {self.retry_count} retry of missing tag in saved marc record"
                     )
                     self.put_folio_records(marc_instance_tags, instance_id)
+                else:
+                    return False
 
         return True
 
--- a/libsys_airflow/plugins/shared/utils.py
+++ b/libsys_airflow/plugins/shared/utils.py
@@ -3,6 +3,7 @@ import json
 import logging
 import pymarc
 import re
+import time
 import urllib
 
 from typing import Union
@@ -83,10 +84,11 @@ def _subject_with_server_name(**kwargs):
 
 
 class FolioAddMarcTags(object):
+    SLEEP = 30
+
     def __init__(self, **kwargs):
         self.httpx_client = httpx.Client()
         self.folio_client = folio_client()
-        self.retry_count = 0
 
     def put_folio_records(self, marc_instance_tags: dict, instance_id: str) -> bool:
         try:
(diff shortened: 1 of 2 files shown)


tests/digital_bookplates/test_add_marc_tags.py
```diff
@@ -170,6 +173,7 @@ def test_put_folio_records_unique_tag(mock_folio_add_marc_tags, caplog):
         r"=979  \\$fABBOTT$bdruid:ws066yy0421$cws066yy0421_00_0001.jp2$dThe The Donald P. Abbott Fund for Marine Invertebrates tag is unique"
         in caplog.text
     )
+    assert "Making 1 retry of missing tag in saved marc record" in caplog.text
```
shelleydoljack, line 176: Can you make the retry a separate test? The test here checks whether or not the tag it was adding was unique and I'd rather not have it conflated with retrying. So I think the mock_httpx_success fixture should have mock_content contain the ABBOTT bookplate data and then create another test with a different success fixture that has the BAILEYT bookplate data. Thanks!

shelleydoljack (changes requested):

tests/digital_bookplates/test_add_marc_tags.py
jgreben: I can make a separate test, but in  the test `test_put_folio_records_unique_tag` it will still be attempting to do a retry, and since we are mocking the httpx.Client irregardless of whether it does a get or a put, the mock will still need to return some content like this so that it does not fail the retry:
```
def mock_httpx_put_success():
    mock_content = b'{"records": [{"parsedRecord": {"content": {"fields": []}}}]}'
    return httpx.Client(
        transport=httpx.MockTransport(
            lambda request: httpx.Response(202, content=mock_content)
        )
    )
```

I tried mocking the get and the put separately but since the method is checking for the client's response code it throws an error because the `put` method does not have a response_code attribute.

So to me it kind of makes sense to test it with the same test, but I will let you decide if you still would like separate tests.

shelleydoljack: I'm mainly concerned about the `__check_retry_put__` function not being explicitly tested so let's add a test for it and leave the test you modified as is.

libsys_airflow/plugins/shared/utils.py
```diff
@@ -126,8 +127,44 @@ def put_folio_records(self, marc_instance_tags: dict, instance_id: str) -> bool:
             logger.info(
                 f"Successfully updated FOLIO Instance {instance_id} with SRS {srs_uuid}"
```
shelleydoljack, line 128: We should move this comment after `__check_retry_put` is called. And maybe modify to comment to "Request to update FOLIO Instance `{instance_id}` with SRS `{srs_uuid}` successfully submitted."

PR #1572, opened 2025-06-24
fix address of discoverySuppress in item

jgreben:
Fixes bug in sal3/caiasoft sql.

### DIFF
--- a/libsys_airflow/plugins/data_exports/sql/sal3_items.sql
+++ b/libsys_airflow/plugins/data_exports/sql/sal3_items.sql
@@ -8,7 +8,7 @@ select
     I.permanentLocationId as item_permanent_location,
     I.temporaryLocationId as item_temporary_location,
     H.permanentLocationId as holdings_permanent_location,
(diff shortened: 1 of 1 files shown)


libsys_airflow/plugins/data_exports/sql/sal3_items.sql
```diff
@@ -8,7 +8,7 @@ select
     I.permanentLocationId as item_permanent_location,
     I.temporaryLocationId as item_temporary_location,
     H.permanentLocationId as holdings_permanent_location,
-    I.discoverySuppress as item_suppressed
+    I.jsonb -> 'discoverySuppress' as item_suppressed
```
shelleydoljack, line 11: When I went looking into discoverySuppress being null for some instances since upgrading to Q CSP8, I saw that item records do not always have a discoverySuppress field as well. I'm not entirely sure how a missing discoverySuppress field would affect what we're doing here.

dlrueda (approved):

PR #1578, opened 2025-07-03
Open course reserves file to truncate.

shelleydoljack:


shelleydoljack:
> Does it not work if you just omitted 165-67 and only open the file for origin on
> 
> https://github.com/sul-dlss/libsys-airflow/blob/8ce233bb211b233ef991825f3e61e81cc66e1fe5/libsys_airflow/plugins/folio/courses.py#L175
> 
> ?
> `'a' (append mode): Opens the file for appending. If the file exists, new content is added to the end of the file without overwriting existing content. If the file does not exist, a new file is created`

Ahh, I suppose we can open it outside the looping through the list of mapped tasks.... I'll try that.

PR #1615, opened 2025-10-15
Use binary mode to calculate file size using FTP

### LINKED ISSUE #1614: VMA error 550 SIZE not allowed in ASCII mode
While testing VMA for retrieving Backstage files I am getting this error https://sul-libsys-airflow-stage.stanford.edu/dags/data_fetcher/grid?run_id=manual__2025-10-14T[…]4T15%3A39%3A04%2B00%3A00&task_id=ftp_download_task

`ftplib.error_perm: 550 SIZE not allowed in ASCII mode` 
Per Jeremy - 
> Doing a quick google search indicates that we need to set the FTP connection mode to binary and not ASCII in order for the SIZE command to work correctly.

Is this related to the [change](https://github.com/sul-dlss/libsys-airflow/pull/1586) made around zero-byte files?

jermnelson:
Fixes #1614, FYI, "TYPE I" stands for images which are binary files.

### DIFF
--- a/libsys_airflow/plugins/vendor/download.py
+++ b/libsys_airflow/plugins/vendor/download.py
@@ -44,6 +44,7 @@ class FTPAdapter:
 
     def get_size(self, filename: str) -> int | None:
         try:
+            self.hook.conn.sendcmd("TYPE I")  # type: ignore
             file_size = self.hook.get_size(filename)
         except ftplib.error_perm as e:
             logger.error(f"Failed to retrieve size for {filename}, {e}")


ahafele:
> > Are there files we download that are non-MARC so sending this command will mess those up?
> 
> I don't think so. Downloading in raw binary would preserve any non-MARC file as-is; the reverse is not true, downloading a binary file in ASCII mode has the potential of the FTP server converting line endings thereby corrupting the file. @ahafele did test other VMA loads with this change (although not sure if there non-MARC files) and those loads worked fine.

I can test some feeder files today

ahafele:
I ran a test to pick up and load an EDI invoice. The file was fetched and sent to data import. It failed in data import I suspect because the orders are not on -stage. I downloaded the same invoice file from prod vma and did a diff and the files are the same so I think it is good to go.

jgreben (dismissed):
Is there anything that can be changed at https://github.com/sul-dlss/libsys-airflow/blob/t1614-ftp-binary/tests/vendor/test_download.py#L160 in order to test that both a binary and non-binary file can be downloaded using this mode?

PR #1657, opened 2025-11-20
Add additional email recipients to vendor load emails

### LINKED ISSUE #1092: VMA - add additional email to reports
Users would like to be able to curate the emails they receive from VMA, e.g. SUL doesn't need to see Law's load reports. Could we add a per interface configuration option to add an email address for reports specific to that interface? This was the pattern in Eloader. User could then unsubscribe to folio-load list and input their email of preference into specific interfaces. Needs to allow for multiple addresses to be input.

jgreben:
Fixes #1092 

- Adds a VendorInterface field to the model for additional emails
- Updates the edit and interface views to display the new VendorInterface field
- Pulls in field data using the vendor uuid and vendor interface uuid and formats the combined emails 
- Update one test to use a mock interface containing the additional emails

### DIFF
--- a/libsys_airflow/plugins/vendor/emails.py
+++ b/libsys_airflow/plugins/vendor/emails.py
@@ -10,7 +10,6 @@ from airflow.providers.postgres.hooks.postgres import PostgresHook
 from sqlalchemy.orm import Session
 from jinja2 import Template
 
-# from libsys_airflow.plugins.vendor_app.database import Session
 from libsys_airflow.plugins.vendor.models import VendorInterface
 from libsys_airflow.plugins.vendor.marc import (
     is_marc,


libsys_airflow/plugins/vendor/emails.py
```diff
@@ -31,6 +32,10 @@ def files_fetched_email_task(downloaded_files: list, kwargs: dict):
     kwargs["vendor_interface_url"] = _vendor_interface_url(
         kwargs["vendor_uuid"], kwargs["vendor_interface_uuid"]
     )
+    kwargs["interface_additional_emails"] = _additional_email_recipients(
+        kwargs["vendor_uuid"], kwargs["vendor_interface_uuid"]
+    )
+
```
shelleydoljack: Is it possible to get the additional email recipients from the vendor_interface_uuid and not rely on an additional kwarg with key "interface_additional_emails"? This task is called from the [data_fetcher_dag](https://github.com/sul-dlss/libsys-airflow/blob/e9703abda7da0bdc417754990cb4572e6bfc24f4/libsys_airflow/dags/vendor/data_fetcher.py#L97) with kwargs that are set in the params task, which essentially gets the params from https://github.com/sul-dlss/libsys-airflow/blob/e9703abda7da0bdc417754990cb4572e6bfc24f4/libsys_airflow/dags/vendor/data_fetcher.py#L41-L56. Those are set in the [data_fetcher_scheduler](https://github.com/sul-dlss/libsys-airflow/blob/e9703abda7da0bdc417754990cb4572e6bfc24f4/libsys_airflow/dags/vendor/data_fetch_scheduler.py#L46-L54) dag. So a lot of code would have to be changed, I think, to make this work.

jgreben: I changed it to use a different class method that does not require the vendor uuid.

PR #1658, opened 2025-11-20
Adds sdr-files to shared directory

jermnelson:
So that `sdr-files` directory persists between deployments.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -23,7 +23,7 @@ set :deploy_to, "/home/libsys/#{fetch(:application)}"
 
 # Default value for linked_dirs is []
 # set :linked_dirs, %w[]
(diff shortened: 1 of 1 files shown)


jgreben (approved):

config/deploy.rb
```diff
@@ -23,7 +23,7 @@ def alembic_dbs
 
 # Default value for linked_dirs is []
 # set :linked_dirs, %w[]
-set :linked_dirs, %w(.aws config vendor-data vendor-keys data-export-files digital-bookplates fix_encumbrances orafin-files logs authorities)
+set :linked_dirs, %w(.aws config vendor-data vendor-keys data-export-files digital-bookplates fix_encumbrances orafin-files logs authorities sdr-files)
```
shelleydoljack: We need to add this to the cron job to clean up files probably.

PR #1669, opened 2025-11-25
Removes unused DAGs.

shelleydoljack:
I'm tired of updating files we aren't using. There's some small possibility that we may be using these DAGs but not any time soon.


ahafele:
is transmit_pod the same as send_pod?

For the Hathi, sharevde, and google - it is all an open question about if/when we would implement them. Nielsen is looking more likely, with work possibly next summer.

dlrueda:
I'd rather keep them around, I suspect within the year we'll resume with at least Nielsen and SVDE. 

I admit I don't remember what Google is for though. And not sure when Hathi might be a thing again.

PR #1677, opened 2025-12-02
Adds branching join task before ftp or http transmission tasks.

shelleydoljack:
Fixes the branching in the data export send dags:
Screenshot 2025-12-02 at 2 50 23 PM


shelleydoljack:
Adding `trigger_rule="all_done"` to the end task means "All upstream tasks are done with their execution" independent of their state. The result is that the skipped no_files_email task doesn't cause the end task's state to be skipped (like in the above screenshot) but rather have a success state, like here:
Screenshot 2025-12-02 at 5 02 40 PM

PR #1689, opened 2025-12-17
FOLIO Authority Records DAG and UI

### LINKED ISSUE #1687: Automate authority record deletions
Monthly we receive 1 or more delete files for Authority records. Authority deletes are not currently handled in DI and not scheduled for Umbrellaleaf or later. Jeremy has a process that has been run in a jupyter notebook that takes a list of 001s and deletes the appropriate authority record. We would like to automate this process so it doesn't have to be run by Jeremy each month. We should take the easiest path since this will eventually be deprecated once we can run the process directly in FOLIO Data Import. 

- User webform that takes input of a csv of MARC 001s
- Includes field for email address of user (following the pattern we have established in other webforms)
- Process finds and deletes the authority record
- Process emails a report of any record that was not deleted 

Related to https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/756 and https://github.com/sul-dlss/FOLIO-Project-Stanford/issues/730

jermnelson:
Fixes #1687 

New UI:
Screenshot 2025-12-17 at 4 41 05 PM

TODO:
- [x] UI for Upload CSV of 001
- [x] UI Unit Tests

### DIFF
--- a/plugins/remove_old_files_cron.py
+++ b/plugins/remove_old_files_cron.py
@@ -4,6 +4,7 @@ cron = CronTab(user=True)
 
 max_age_days = 365
 
+
 data_export_files = cron.new(
     command=f'find /home/libsys/libsys-airflow/shared/data-export-files -type f -mtime +{max_age_days} -delete'
 )
@@ -52,5 +53,12 @@ logs_files = cron.new(
 logs_files.dow.on('SUN')
 logs_files.hour.on(0)
 logs_files.minute.on(40)
+
+authority_files = cron.new(
+    command=f"find /home/libsys/libsys-airflow/shared/authorities -type f -mtime +{max_age_days} -delete"
+)
+authority_files.dow.on('SUN')
+authority_files.hour.on(0)
+authority_files.minute.on(45)
 
 cron.write()


libsys_airflow/plugins/authority_control/apps/deletes_upload_view.py
```diff
@@ -0,0 +1,67 @@
+import datetime
+import pathlib
+
+import pandas as pd
+
+from airflow.models import DagBag
+from airflow.utils.state import State
```
shelleydoljack: Could you change this to use TriggerDagRunOperator instead of DagBag? The DagBag is not available in Airflow 3. You can see how I changed stuff like this in this commit: https://github.com/sul-dlss/libsys-airflow/pull/1625/commits/df8b311bb4f4a406cc49621c04ad41711c9a333d

libsys_airflow/plugins/authority_control/apps/deletes_upload_view.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class AuthorityRecordsDeleteUploadView(AppBuilderBaseView):
+
+    default_view = "authorities_delete_home"
+    route_base = "/authorities_deletes"
+
+    def _save_deletes_csv(self, deletes_df: pd.DataFrame, filename: str) -> str:
+        authority_uploads_path = pathlib.Path("/opt/airflow/authorities/uploads")
+        authority_uploads_path.mkdir(parents=True, exist_ok=True)
```
shelleydoljack: Is this path in the shared directory so as to not be overwritten every time we deploy?

libsys_airflow/plugins/authority_control/apps/deletes_upload_view.py
shelleydoljack: In that commit, I had added a function to `libsys_airflow/plugins/shared/utils.py` to return an execution_date for using in dags:
```
from datetime import datetime, timezone

def execution_date() -> str:
    return datetime.now(timezone.utc).isoformat()

```

jermnelson: Hi Shelley, I'm not sure how to use the `TriggerDagRunOperator` outside of a DAG when we're trying to trigger a DAG run from the UI?

shelleydoljack: Does something like this https://github.com/sul-dlss/libsys-airflow/blob/df8b311bb4f4a406cc49621c04ad41711c9a333d/libsys_airflow/plugins/data_exports/apps/data_export_upload_view.py#L51-L69 not work?

shelleydoljack (approved):

Issue #1621, opened 2025-10-22
Update plugins to use FAB provider for Airflow 3 compatibility


shelleydoljack:
Illustrative example of a plugin for Airflow 3: https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/plugins.html#example

> install the FAB provider and tweak the code following Airflow 3 migration guide. This compatibility layer allows you to continue using your existing Flask Appbuilder views, Flask Blueprints and Flask Appbuilder menu items.
https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/plugins.html#flask-appbuilder-and-flask-blueprints-in-airflow-3

shelleydoljack:
This ticket is related to #1627 . We likely want to use [FAB auth manager](https://airflow.apache.org/docs/apache-airflow-providers-fab/stable/auth-manager/index.html).

shelleydoljack:
Fixed in https://github.com/sul-dlss/libsys-airflow/pull/1705

shelleydoljack:
Actually, I reopened it b/c I'm not sure if there is more to this than just adding FAB auth provider in the airflow.cfg.

shelleydoljack:
I'm going to close this ticket because our Flask appbuilder_views still work in Airflow 3. However, any new plugins we create, we should go back and refactor to use the Airflow 3 plugins. See docs: 

> Airflow 2 supported Flask Appbuilder views (appbuilder_views), Flask AppBuilder menu items (appbuilder_menu_items), and Flask Blueprints (flask_blueprints) in plugins. These have been superseded in Airflow 3 by External Views (external_views), Fast API apps (fastapi_apps), FastAPI middlewares (fastapi_root_middlewares) and React apps (react_apps) that allow extended functionality and better integration with the Airflow UI.
https://airflow.apache.org/docs/apache-airflow/3.1.8/administration-and-deployment/plugins.html#flask-appbuilder-and-flask-blueprints-in-airflow-3

PR #1705, opened 2026-02-12
Upgrade to Airflow 3.1.7

jermnelson:
Updates code and tests to run Airflow 3.1.7. Still need to figure out how to trigger a DAG run from code not running in a DAG.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -49,7 +49,6 @@ x-airflow-common:
   # In order to add custom dependencies or upgrade provider packages you can use your extended image.
   # Comment the image line, place your Dockerfile in the directory where you placed the docker-compose.yaml
   # and uncomment the "build" line below, Then run `docker-compose build` to build the images.
-  # image: ${AIRFLOW_IMAGE_NAME:-apache/airflow:3.1.7}
   build: .
   environment:
     &airflow-common-env
--- a/compose.prod.yaml
+++ b/compose.prod.yaml
@@ -50,7 +50,6 @@ x-airflow-common:
   # In order to add custom dependencies or upgrade provider packages you can use your extended image.
   # Comment the image line, place your Dockerfile in the directory where you placed the docker-compose.yaml
   # and uncomment the "build" line below, Then run `docker-compose build` to build the images.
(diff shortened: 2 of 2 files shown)


airflow.cfg
```diff
@@ -1,6 +1,10 @@
 [core]
+auth_manager = airflow.providers.fab.auth_manager.fab_auth_manager.FabAuthManager
```
shelleydoljack: Does this address this issue? https://github.com/sul-dlss/libsys-airflow/issues/1621

Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM apache/airflow:3.1.0-python3.10
+FROM apache/airflow:3.1.7-python3.12
```
shelleydoljack: We can't get python 3.12 onto the libsys-airflow dev server. At least that was what this convo was about https://stanfordlib.slack.com/archives/C0RK5EM9N/p1760633892647619

compose.yaml
```diff
@@ -49,7 +49,7 @@ x-airflow-common:
   # In order to add custom dependencies or upgrade provider packages you can use your extended image.
   # Comment the image line, place your Dockerfile in the directory where you placed the docker-compose.yaml
   # and uncomment the "build" line below, Then run `docker-compose build` to build the images.
-  # image: ${AIRFLOW_IMAGE_NAME:-apache/airflow:2.8.3}
+  # image: ${AIRFLOW_IMAGE_NAME:-apache/airflow:3.1.7}
```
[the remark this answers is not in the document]
jermnelson: Good point; removing.

airflow.cfg
jermnelson: Yes, I think it does. When bringing up the Airflow locally, all of our custom plugin views are showing up.

Dockerfile
jermnelson: After the Staff meeting, we'll be going with option 1  @tallenaz

tallenaz: Sounds, good, I'll get started on option 1.

PR #1731, opened 2026-04-14
Adds API server JWT and secret key env vars to docker compose files.

### LINKED ISSUE #1627: Add jwt_secret for API authentication
See docs about [jwt_secret](https://airflow.apache.org/docs/apache-airflow/3.1.0/configurations-ref.html#jwt-secret)
We probably want to pull in the secret from Vault.
When not set, on startup the airflow-api-server logs show:
```
2025-10-22T21:11:43.759158Z [warning  ] `api_auth/jwt_secret` was empty, using a generated one for now. Please set this in your config [airflow.api_fastapi.auth.tokens] loc=tokens.py:529
```

shelleydoljack:
Closes #1627 

I think we should manage the values for `AIRFLOW__API_AUTH__JWT_SECRET` and `AIRFLOW__API_AUTH__JWT_ISSUER` in vault so that we can ensure the deployment grabs the correct values and not get into figuring which shell's env vars are involved. If so, I will add a follow-on PR to https://github.com/sul-dlss/puppet/pull/13598 to add the vault lookups.


jermnelson (approved):

compose.prod.yaml
```diff
@@ -61,6 +61,11 @@ x-airflow-common:
     AIRFLOW__CORE__FERNET_KEY: ${AIRFLOW__CORE__FERNET_KEY:-}
     AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION: 'true'
     AIRFLOW__CORE__LOAD_EXAMPLES: 'false'
+    AIRFLOW__CORE__EXECUTION_API_SERVER_URL: 'http://airflow-apiserver:8080/execution/'
```
jgreben: Where in the docs does it say to add these particular env vars?

compose.prod.yaml
```diff
@@ -61,6 +61,11 @@ x-airflow-common:
...
+    AIRFLOW__API_AUTH__JWT_SECRET: ${AIRFLOW__API_AUTH__JWT_SECRET:-airflow_jwt_secret}
+    AIRFLOW__API_AUTH__JWT_ISSUER: ${AIRFLOW__API_AUTH__JWT_ISSUER:-airflow}
+    AIRFLOW__API__SECRET_KEY: ${AIRFLOW__API__SECRET_KEY}
+    AIRFLOW__API__BASE_URL: ${AIRFLOW__API__BASE_URL}
```
jgreben, line 68: Where in the docs does it say to add these particular env vars? In the issue you only referenced [jwt_secret](https://airflow.apache.org/docs/apache-airflow/3.1.0/configurations-ref.html#jwt-secret)

shelleydoljack: `AIRFLOW__API__SECRET_KEY` and `AIRFLOW__API__BASE_URL` are replacements for the webserver ones for Airflow 2.
I believe the env vars related to API_AUTH__JWT are required now because Airflow 3 makes the workers and triggerers use the api server.

PR #1732, opened 2026-04-15
Change marc transformation to use 950 instead of 999

### LINKED ISSUE #1730: Change POD export to use 950 MARC tag
POD currently uses the 999 and we want to change it to 950 (ahead of google books export). 

We need to 

1. Pause POD dags
2. Send new mapping to reshare rep(s)
2. Update code to use 950 for full and ongoing POD exports
4. Send new full dump to POD

jgreben:
Fixes #1730 assuming all of the subfields of the 950 are the same as the 999, including:

Holdings subfield = folio holdings field
h = holdingsTypeId
l = permanentLocationId
w = callNumberTypeId
r = illPolicyId
a = callNumber

Item subfield = folio items field
i = itemBarcode
t = materialTypeId
e = effectiveLocationId
j = numberOfPieces
a = enumeration

Keep in draft pending step 3 from #1730:

- [ ] Pause POD dags
- [x] Send new mapping to reshare rep(s)
- [ ] Update code to use 950 for full and ongoing POD exports (merge this PR)
- [ ] Send new full dump to POD


dlrueda:
New mapping has been sent to Reshare

dlrueda:
To send a new full dump to POD (from Cory):
create a new stream, send a full dump to it, then start sending updates to it. Once that's all set up let me know and I'll switch it over to the default.

jermnelson (approved):

shelleydoljack:
> To send a new full dump to POD (from Cory): create a new stream, send a full dump to it, then start sending updates to it. Once that's all set up let me know and I'll switch it over to the default.

I logged into pod.stanford.edu to see if I could create a new stream, because I thought we would do that manually, but I don't see a way to do that with my user access. https://pod.stanford.edu/organizations/stanford/streams

Also noticed that we'll need to update the info on this page too: https://pod.stanford.edu/organizations/stanford/provider_details

libsys_airflow/dags/data_exports/full_dump_retrieval.py
```diff
@@ -88,7 +94,7 @@
             "marc-files",
             type="string",
             description="The S3 marc file path to deposit the MARC records. CC0 or marc-files.",
```
shelleydoljack: We should update the description if the enum values have changed.

shelleydoljack (approved):
Nice. I hope the SQL does what we expect b/c I don't want to write it. 😀 
Just one comment to update a description field.

libsys_airflow/dags/data_exports/full_dump_retrieval.py
```diff
@@ -87,8 +93,8 @@
         "marc_file_dir": Param(
             "marc-files",
             type="string",
-            description="The S3 marc file path to deposit the MARC records. CC0 or marc-files.",
-            enum=["marc-files", "CC0"],
+            description="The S3 marc file path to deposit the MARC records for POD, CC0 or Google Books.",
+            enum=["pod-files", "CC0", "google-files"],
```
shelleydoljack, line 97: Actually, if you change the filepath in S3 where these are written to, then you need to update where the full dump transmission tasks pickup files: https://github.com/sul-dlss/libsys-airflow/blob/1dc6c323da4684f917f2b6e3bdb26952f610cfd9/libsys_airflow/plugins/data_exports/transmission_tasks.py#L43

shelleydoljack: Honestly, I'd just leave it as `marc-files` instead of making it vendor-specific with "pod" in the name. Then we won't need to tweak anything with the transmission of these things.

shelleydoljack:
And now I'm remembering something about the streams. The transmission of the files will automatically go to the default stream. So we actually want Cory to create a new stream and set it to the default before we transmit files in order for them to go to the correct stream.

dlrueda:
Not sure if us needing a default stream to send the full dump to works with what Cory said he expects.  Can someone coordinate what to do directly with Cory?  (I'll try to catch him during Anchor Week, but I'm also not the one who will kick off the dag in prod once we've verified the numbers look ok)

libsys_airflow/dags/data_exports/full_dump_retrieval.py
jgreben: https://github.com/sul-dlss/libsys-airflow/blob/3ea84191d8f5029093e83765757fdf2042a04620/libsys_airflow/plugins/data_exports/transmission_tasks.py#L44

jgreben: Since your prior comment I already did the work to parameterize the s3 files dir for the vendors.

libsys_airflow/dags/data_exports/full_dump_transmission.py
```diff
@@ -37,7 +37,16 @@ def retrieve_params(**kwargs):
     """
     params = kwargs.get("params", {})
     conn_id = params["vendor"]
-    return {"conn_id": conn_id}
+
+    match conn_id:
+        case "pod":
+            marc_file_dir = "marc-files"
```
shelleydoljack: This should be `marc_file_dir = "pod-files"` based on https://github.com/sul-dlss/libsys-airflow/blob/3ea84191d8f5029093e83765757fdf2042a04620/libsys_airflow/dags/data_exports/full_dump_retrieval.py#L97. And if "marc-files" is the default case, then line 97 needs to include marc-files as a value of the enum. Or just revert commit  https://github.com/sul-dlss/libsys-airflow/pull/1732/commits/3ea84191d8f5029093e83765757fdf2042a04620 and change line 97 enum values to `enum=["marc-files", "CC0", "google-files"]`

libsys_airflow/dags/data_exports/full_dump_transmission.py
```diff
@@ -37,7 +37,16 @@ def retrieve_params(**kwargs):
...
     conn_id = params["vendor"]
-    return {"conn_id": conn_id}
+
+    match conn_id:
+        case "pod":
+            marc_file_dir = "pod-files"
+        case "google-books":
+            marc_file_dir = "google-files"
+        case _:
+            marc_file_dir = "marc-files"
```
shelleydoljack: The default case will never occur since the params passed are never "marc-files" based on the enum list.

jgreben: The enum is `enum=["pod", "sharevde", "backstage", "google-books"]` so it's possible to send something besides `pod` or `google-books` hence a default (as "defensive" coding) even though we will probably never kick off a dag with those other vendors.

shelleydoljack: I meant the enum list for the full_dump retrieval, where we specify the folder that the files are written. You changed it from marc-files to pod-files, so there will never be a directory folder in S3 called "marc-files" for a transmission dag to pickup from.
And in fact, looking back at the Param, we have the default set to "marc-files" with an enumerated list that doesn't include "marc-files". Isn't that problematic in airflow?
```
 "marc_file_dir": Param(
            "marc-files",
            type="string",
            description="The S3 marc file path to deposit the MARC records for POD, CC0 or Google Books.",
            enum=["pod-files", "CC0", "google-files"],
```
Looking at the gather_files task that all the transmission DAGs use, we default files_dir to "marc-files" for full-dump if there is not a files_dir kwarg sent. It just seems really fraught with potential bugs to have changed the file directory that the full dump writes files to. We wrote it to be like this. 🤷
```
marc_filepath = Path(airflow) / f"data-export-files/{vendor}/marc-files/"
    file_glob_pattern = vendor_fileformat_spec(vendor)
    if vendor == "full-dump":
        files_dir = kwargs.get("files_dir", 'marc-files')
        marc_filepath = S3Path(f"/{bucket}/data-export-files/{vendor}/{files_dir}/")
    marc_filelist = []
    for f in marc_filepath.glob(file_glob_pattern):
        if f.stat().st_size in [0, 112]:
            continue
        marc_filelist.append(str(f))
```
@ahafele is the specification for a full dump of pod files somehow different than for other potential full dump consumers, that they should be written to a directory specific for pod? We had thought a full dump would always be the same across all consumers, but now we have CC0 and google. Are there more that will be special snowflakes?

jgreben: In this case I think "perfect" is going to be the enemy of the good. I updated the vendor params so that pod-files is the default. I am going to run a full pod dump on test today. Unless there are major bugs that surface can we just go ahead and get this running so that pod and google can get their data?

shelleydoljack: Sure. Did you want to merge it now or wait until after a test run to see if any more bugs surface?

jgreben: I'm going to deploy and run it on airflow-dev.

jgreben:
https://sul-libsys-airflow-dev.stanford.edu/dags/select_all_records/grid?dag_run_id=manual__2026-04-27T12%3A23%3A37-07%3A00

and

https://sul-libsys-airflow-dev.stanford.edu/dags/send_all_records/grid?dag_run_id=manual__2026-04-28T11%3A06%3A39-07%3A00&task_id=gather_files_task&tab=logs

libsys_airflow/dags/data_exports/pod_selections.py
```diff
@@ -54,14 +54,14 @@ def missing_marc_records_email(**kwargs):
...
-def compress_marc_files(marc_file_list: dict):
+def compress_marc_files(marc_file_list: dict, vendor: str):
     marc_files = (
         marc_file_list['new'] + marc_file_list['updates'] + marc_file_list['deletes']
     )
     for marc_file in marc_files:
         stem = pathlib.Path(marc_file).suffix
         xml = marc_file.replace(stem, '.xml')
-        zip_marc_file(xml)
+        zip_marc_file(xml, vendor="pod")
```
shelleydoljack, line 64: It seems weird to have to pass a vendor param for a DAG that is specific to the vendor. But we probably do that elsewhere.

libsys_airflow/plugins/data_exports/transmission_tasks.py
```diff
@@ -34,14 +34,30 @@ def gather_files_task(**kwargs) -> dict:
...
+            case "pod":
+                marc_files_dir = "pod-files"
+            case "google":
+                marc_files_dir = "google-files"
+            case _:
+                marc_files_dir = "marc-files"
+
+        logger.info(f"Full dump with files_dir {marc_files_dir}")
+        marc_filepath = S3Path(
+            f"/{bucket}/data-export-files/full-dump/{marc_files_dir}/"
```
shelleydoljack, line 57: I see you introduced a new directory here. I guess at least the archive_transmitted_data_task is skipped for full-dumps, otherwise the code that's already really gnarly to read gets even more so.

shelleydoljack (approved):

PR #1739, opened 2026-04-22
Adds abiliity to get access tokens for the API user

shelleydoljack:
Adding the apache-airflow-client made a change to the poetry.lock file. To reconcile your env locally, I would suggest making sure your python 3.12 poetry env is activitated, then do `poetry sync` (not install).


jermnelson (approved):

shelleydoljack:
This is how I used the airflow client in the connections.py that I'm working on:
```
from libsys_airflow.plugins.shared.airflow_client import (
    api_client,
    client_configuration,
    get_access_token,
)
    configuration = client_configuration
    # set access token for client header
    configuration.access_token = get_access_token(
        host=client_configuration.host,
        username=client_configuration.username,
        password=client_configuration.password,
    )
    client = api_client(configuration)
    api_instance = airflow_client.client.ConnectionApi(client)
```

shelleydoljack:
FYI https://github.com/sul-dlss/puppet/pull/13630

PR #1741, opened 2026-04-28
Removes DB access from create_connection_task

### LINKED ISSUE #1622: Change code that uses direct database access
> In Airflow 3 operators can not access the Airflow metadata database directly using database sessions. If you have custom operators, review the code to make sure there are no direct db access. You can follow examples in https://github.com/apache/airflow/issues/49187 to find how to modify your code if needed.

`libsys-airflow/libsys_airflow/plugins/airflow/connections.py` needs to be changed.

shelleydoljack:
Closes #1622 

I believe this is the only place where we try to interact with the Airlfow "meta" database.

I was able to run the data_fetcher DAG locally with folio-stage-api and I'm hoping how I've configured the compose files, we don't run into "is it supposed to be localhost or the docker service name or etc????" when it comes to figuring out the AIRFLOW__API__BASE_URL. (it didn't work with localhost:8080 b/c that's not the name of the docker container service)


libsys_airflow/plugins/shared/airflow_api_client.py
```diff
@@ -36,7 +28,15 @@ def get_access_token(
     return response_success.access_token
 
 
-def api_client(
-    configuration: airflow_client.client.Configuration,
-) -> airflow_client.client.ApiClient:
+def api_client() -> airflow_client.client.ApiClient:
```
jermnelson, line 31: I like how the api_client is cleaner and easier to use with configuration internal to the function.

jermnelson (approved):
Looks good; finding the Airflow API client was a good choice for easier migration to Airflow 3.

jgreben (approved):

PR #1743, opened 2026-04-29
Fixes API base url settings

shelleydoljack:
Related to this puppet PR: https://github.com/sul-dlss/puppet/pull/13650

Sets the `AIRFLOW__API__BASE_URL` to the airflow-apiserver docker container service, which the other airflow containers use when interacting with the Airflow 3 API. Having airflow-apiserver service running on ports `3000:8080` aligns with the server's proxy_pass settings:
https://github.com/sul-dlss/puppet/blob/9a7b049371e128e3a355db47fa81a92d4d8fa4ce/hieradata/node/sul-libsys-airflow-dev-up.stanford.edu.yaml#L182

### DIFF
--- a/compose.prod.yaml
+++ b/compose.prod.yaml
@@ -65,7 +65,7 @@ x-airflow-common:
     AIRFLOW__API_AUTH__JWT_SECRET: ${AIRFLOW__API_AUTH__JWT_SECRET:-airflow_jwt_secret}
     AIRFLOW__API_AUTH__JWT_ISSUER: ${AIRFLOW__API__BASE_URL:-airflow}
     AIRFLOW__API__SECRET_KEY: ${AIRFLOW__API__SECRET_KEY}
-    AIRFLOW__API__BASE_URL: ${AIRFLOW__API__BASE_URL}
+    AIRFLOW__API__BASE_URL: 'http://airflow-apiserver:8080'
     # yamllint disable rule:line-length
     # Use simple http server on scheduler for health checks
     # See https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/logging-monitoring/check-health.html#scheduler-health-check-server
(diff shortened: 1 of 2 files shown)


compose.prod.yaml
```diff
@@ -65,7 +65,7 @@ x-airflow-common:
     AIRFLOW__API_AUTH__JWT_SECRET: ${AIRFLOW__API_AUTH__JWT_SECRET:-airflow_jwt_secret}
     AIRFLOW__API_AUTH__JWT_ISSUER: ${AIRFLOW__API__BASE_URL:-airflow}
     AIRFLOW__API__SECRET_KEY: ${AIRFLOW__API__SECRET_KEY}
-    AIRFLOW__API__BASE_URL: ${AIRFLOW__API__BASE_URL}
+    AIRFLOW__API__BASE_URL: 'http://airflow-apiserver:8080'
```
shelleydoljack: This is where we define the docker containers' `os.getenv` environment variables so that the airflow_api_client can use this value. But for the host server configuration, I'm still not sure how we need to define localhost:3000 is `sul-libsys-airflow-dev-up`.

shelleydoljack: Forget this; it was a bad idea.

shelleydoljack:
I reverted the puppet change with https://github.com/sul-dlss/puppet/pull/13653 and hardcoded host for the airflow_api_client to be the docker container service. I think all the airflow containers don't have an issue with the env var `AIRFLOW__API__BASE_URL` being set to the server hostname because they have their own env var to interact with the API: `AIRFLOW__CORE__EXECUTION_API_SERVER_URL` Plus, by having the API base url set to the external FQDN, we could use the airflow API outside of the docker network if we wanted to.

jermnelson (approved):

jgreben (approved):

PR #1748, opened 2026-05-01
Use REST airflow_client to Trigger Circ Rules DAG

jermnelson:
Along with other changes to support correct routing in Plugin for batch and reports

**NOTE**: Discovered issue with trying to download bulk report or reference data is that downloads are prevented in the iframe that the Circ Rules Plugin runs in. Doing a right-mouse click to open the report another tab allows you download the report or csv as a workaround.

### DIFF
--- a/libsys_airflow/plugins/folio/apps/circ_rules_tester_view.py
+++ b/libsys_airflow/plugins/folio/apps/circ_rules_tester_view.py
@@ -14,10 +14,6 @@ from libsys_airflow.plugins.shared.airflow_api_client import api_client
 
 CIRC_HOME = "/opt/airflow/circ"
 
-import logging
-
-logger = logging.getLogger(__name__)
-
 
 class CircRulesTester(AppBuilderBaseView):
     default_view = "circ_home"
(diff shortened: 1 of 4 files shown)


shelleydoljack (approved):

shelleydoljack:
> Along with other changes to support correct routing in Plugin for batch and reports
> 
> **NOTE**: Discovered issue with trying to download bulk report or reference data is that downloads are prevented in the iframe that the Circ Rules Plugin runs in. Doing a right-mouse click to open the report another tab allows you download the report or csv as a workaround.

Do you know if this is because we are using the legacy view? In other words, if we refactor our plugins to use, for instance, FastAPI apps, would this be an issue?

jermnelson:
> > Along with other changes to support correct routing in Plugin for batch and reports
> > **NOTE**: Discovered issue with trying to download bulk report or reference data is that downloads are prevented in the iframe that the Circ Rules Plugin runs in. Doing a right-mouse click to open the report another tab allows you download the report or csv as a workaround.
> 
> Do you know if this is because we are using the legacy view? In other words, if we refactor our plugins to use, for instance, FastAPI apps, would this be an issue?

I don't think so; I believe the iframe is approach is to support legacy plugins but I'll find out.

PR #1751, opened 2026-05-04
Version bump for Airflow 3 upgrade

jermnelson:
Because Airflow 3 is a definite change from our Airflow 2.x code base.


jermnelson:
Extra commits that shouldn't be part of this PR.

PR #1752, opened 2026-05-04
Airflow3 github actions and linting patches

shelleydoljack:


jermnelson (approved):
Since you're touching these configuration files, do you want to up the version in `pyproject.toml` to 2.0.0 as well or wait until we merge the airflow3-upgrade branch into main?

PR #1757, opened 2026-05-05
Refactors DAG979Sensor

### LINKED ISSUE #1640: Rewrite DAG979Sensor to use public API
The DAG979Sensor pokes the DagRun model to get the state and conf of digital_bookplate_979 dagruns it is monitoring until the state is success or failed. The DagRun model is not available in Airflow 3. An alternative would be to use the public Airflow API, specifically the [Get Dag Run](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html#operation/get_dag_run) endpoint, which returns a JSON response that contains the info needed for the poll_979s_dags task:
```
{
  "dag_run_id": "string",
  "dag_id": "string",
  "logical_date": "2019-08-24T14:15:22Z",
  "queued_at": "2019-08-24T14:15:22Z",
  "start_date": "2019-08-24T14:15:22Z",
  "end_date": "2019-08-24T14:15:22Z",
  "duration": 0,
  "data_interval_start": "2019-08-24T14:15:22Z",
  "data_interval_end": "2019-08-24T14:15:22Z",
  "run_after": "2019-08-24T14:15:22Z",
  "last_scheduling_decision": "2019-08-24T14:15:22Z",
  "run_type": "backfill",
  "state": "queued",
  "triggered_by": "cli",
  "triggering_user_name": "string",
  "conf": {},
  "note": "string",
  "dag_versions": [
    {
      "id": "497f6eca-6276-4993-bfeb-53cbbbba6f08",
      "version_number": 0,
      "dag_id": "string",
      "bundle_name": "string",
      "bundle_version": "string",
      "created_at": "2019-08-24T14:15:22Z",
      "dag_display_name": "string",
      "bundle_url": "string"
    }
  ],
  "bundle_version": "string",
  "dag_display_name": "string"
}
```
In order to use the API, we will need to configure our JWT using FAB auth manager (since our plugins use the FAB provider). See https://airflow.apache.org/docs/apache-airflow/stable/security/api.html#public-api

shelleydoljack:
Closes #1640

### DIFF
--- a/libsys_airflow/plugins/digital_bookplates/dag_979_sensor.py
+++ b/libsys_airflow/plugins/digital_bookplates/dag_979_sensor.py
@@ -35,11 +35,11 @@ class DAG979Sensor(BaseSensorOperator):
                     logger.warning(
                         f"No dag run found for digital_bookplate_979 with run ID {dag_run_id}"
                     )
-                    continue
                 else:
                     logger.info(
                         f"Exception when calling DagRunApi for {dag_run_id}: {e}"
                     )
+                continue
 
             state = api_response.state.name.lower()
             self.dag_runs[dag_run_id]['state'] = state
@@ -63,7 +63,7 @@ class DAG979Sensor(BaseSensorOperator):
 
                 self.dag_runs[dag_run_id]['instances'] = instances
         poke_result = all(
-            [val['state'] in ['success', 'failed'] for val in self.dag_runs.values()]
+            [val['state'] in ['success', 'failed'] for val in self.dag_runs.values() if val['state'] is not None]
         )
         logger.info(f"Result of polling DAGs {poke_result}")
         return poke_result
--- a/libsys_airflow/plugins/digital_bookplates/dag_979_sensor.py
+++ b/libsys_airflow/plugins/digital_bookplates/dag_979_sensor.py
@@ -42,11 +42,11 @@ class DAG979Sensor(BaseSensorOperator):
                 continue
 
             state = api_response.state.name.lower()
-            self.dag_runs[dag_run_id]['state'] = state
-            self.dag_runs[dag_run_id]['url'] = dag_run_response_url(
-                dag_run=api_response
-            )
             if state in ['success', 'failed']:
+                self.dag_runs[dag_run_id]['state'] = state
+                self.dag_runs[dag_run_id]['url'] = dag_run_response_url(
+                    dag_run=api_response
+                )
                 instances = []
                 for instance, bookplates in api_response.conf[
                     'druids_for_instance_id'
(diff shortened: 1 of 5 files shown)


libsys_airflow/plugins/digital_bookplates/dag_979_sensor.py
```diff
@@ -18,18 +23,32 @@ def __init__(self, dag_runs: list, **kwargs):
...
+            except ApiException as e:
+                if e.status == 404:
+                    logger.warning(
+                        f"No dag run found for digital_bookplate_979 with run ID {dag_run_id}"
+                    )
+                    continue
+                else:
+                    logger.info(
+                        f"Exception when calling DagRunApi for {dag_run_id}: {e}"
+                    )
```
jermnelson, line 41: Should the loop continue after logging the general exception?

jermnelson (approved):
Approving with question about loop

libsys_airflow/plugins/digital_bookplates/dag_979_sensor.py
shelleydoljack: Probably. Let me see if I can mock a fixture to test that. I'm guessing it won't b/c I don't have a `continue` statement?

shelleydoljack: I went down a bit of rabbit hole with this looping. In the current implementation of the class, if `DagRun.find` doesn't find a dag run, it checks the next dag_run_id but the poke_result will remain False if one of the states of the dag runs is success and the other still None. Here is a snippet of the python debugger when running the test where I've mocked a success and an ApiException:
```
(Pdb) self.dag_runs.values()
dict_values([{'state': 'success', 'instances': [{'uuid': 'd55f7f1b-9512-452c-98ff-5e2be9dcdb16', 'funds': []}], 'url': 'http://localhost:8080/dags/digital_bookplate_979/runs/manual__2024-10-17'}, {'state': None, 'instances': []}])
(Pdb) self.dag_runs
{'manual__2024-10-17': {'state': 'success', 'instances': [{'uuid': 'd55f7f1b-9512-452c-98ff-5e2be9dcdb16', 'funds': []}], 'url': 'http://localhost:8080/dags/digital_bookplate_979/runs/manual__2024-10-17'}, 'manual__2024-10-18': {'state': None, 'instances': []}}
(Pdb) poke_result = all([val['state'] in ['success', 'failed'] for val in self.dag_runs.values()])
(Pdb) poke_result
False
(Pdb) poke_result = all([val['state'] in ['success', 'failed'] for val in self.dag_runs.values() if val['state'] is not None ])
(Pdb) poke_result
True
```
If we change poke_result to include ` if val['state'] is not None`, then we will get True, which I think is what we want. This would mean that poke_result would never return False and I have to change the assertion for test_dag_979_sensor_no_dags. Does this sound right?

shelleydoljack: @jermnelson Let me know if I should revert https://github.com/sul-dlss/libsys-airflow/pull/1757/commits/4c684ae8be63027ee373d68f905e5efacfa3060b

shelleydoljack: 🤔 actually, I see that changing poke_result to never return False will mean that dag runs still in running or queued states won't get re-polled since polling relies on the poke function to be False until it is True. Let me revert the addition I made to poke_result.

PR #1760, opened 2026-05-05
Use airflow client to trigger Boundwith DAG run; add tests

### LINKED ISSUE #1754: Boundwiths Plugin
Use airflow client to trigger DAG run at  https://github.com/sul-dlss/libsys-airflow/blob/b91569b665814221790dd37aff07955f5a669f0a/libsys_airflow/plugins/boundwith/boundwith_view.py#L15-L34

jgreben:
Fixes #1754

### DIFF
--- a/libsys_airflow/plugins/boundwith/boundwith_view.py
+++ b/libsys_airflow/plugins/boundwith/boundwith_view.py
@@ -4,8 +4,8 @@ import pandas as pd
 
 from flask import flash, request
 
-from airflow.providers.standard.operators.trigger_dagrun import TriggerDagRunOperator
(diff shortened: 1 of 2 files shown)


jermnelson (approved):

shelleydoljack (changes requested):
The `trigger_bw_dag` function previously returned the run_id and logical_date but this just returns the run ID. It seems like that is problematic for this to actually work. But looking in prod, the last time this DAG was run was in 2024. 🤷

jgreben:
@shelleydoljack I removed `execution_date` from the return of the function and from the param sent to the view.

shelleydoljack (approved):
🎉

PR #1808, opened 2026-06-02
ap_payment_report dag patches

shelleydoljack:
The ap_payment_report dag run in prod after the upgrade failed because the XCOM for retrieve_invoice_task returns a dictionary and not a list. See task instance https://sul-libsys-airflow-prod.stanford.edu/dags/ap_payment_report/runs/manual__2026-06-01T19:05:17.395270+00:00/tasks/update-folio.update_vouchers_task/mapped/0?try_number=2

I did a "hotfix" on the server to remove the `[0]` list index and reran the tasks and they succeeded.
Another issue was that it also emailed that it failed to update the invoices to status = paid, but it actually didn't fail to update. I think the change in [commit](https://github.com/sul-dlss/libsys-airflow/commit/6832409992090abfa9049d2cf92a7fbf8174e8f2) by returning a dictionar or None will help. Running locally against folio-test, I get:
 
Screenshot 2026-06-01 at 5 56 44 PM

Here is the dag run that shows the one invoice that failed to update went to the email_invoice_errors_task (which I marked succeeded b/c it was failing on airflow.utils.email.send_email from localhost) and the other tasks got marked skipped:
Screenshot 2026-06-01 at 6 52 46 PM

### DIFF
--- a/libsys_airflow/plugins/orafin/reports.py
+++ b/libsys_airflow/plugins/orafin/reports.py
@@ -236,7 +236,7 @@ def update_voucher(
     """
     row = task_instance.xcom_pull(
         task_ids="retrieve_invoice_task", key=voucher["invoiceId"]
-    )[0]
+    )
     voucher["status"] = "Paid"  # should already be Paid when update_invoice task ran
     voucher["disbursementAmount"] = row["AmountPaid"]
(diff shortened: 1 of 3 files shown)


jermnelson (approved):
Thanks for figuring out this error.

PR #1812, opened 2026-06-04
Use file of hrids to dump marc files

jgreben:
Use a file of HRIDs as the sql for google books source records.
Fixes Airflow3 Issue: Don't use runtime-varying value as argument in Dag constructor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,5 +20,3 @@ vendor-keys/
 digital-bookplates/
 authorities/
 sdr-files/
-libsys_airflow/plugins/data_exports/sql/*
-!libsys_airflow/plugins/data_exports/sql/**/*.sql
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ vendor-keys/
 digital-bookplates/
 authorities/
 sdr-files/
+libsys_airflow/plugins/data_exports/sql/*
+!libsys_airflow/plugins/data_exports/sql/**/*.sql
--- a/libsys_airflow/plugins/data_exports/marc/exporter.py
+++ b/libsys_airflow/plugins/data_exports/marc/exporter.py
@@ -210,7 +210,6 @@ class Exporter(object):
         logger.info(f"Writing to directory: {directory}")
         directory.mkdir(parents=True, exist_ok=True)
         marc_file = directory / f"{marc_file_name}.mrc"
-        marc_file.touch()
 
         with marc_file.open(mode) as fo:
             marc_writer = marcWriter(fo)
--- a/libsys_airflow/plugins/data_exports/sql/hrids_mat_view.sql
+++ b/libsys_airflow/plugins/data_exports/sql/hrids_mat_view.sql
@@ -1,11 +1,11 @@
 -- Materialized view built from flat lists of FOLIO instance hrids.
--- The COPY commands below require the txt files to exist on the DB server filesystem
--- and the connecting user to have pg_read_server_files privilege (or be a superuser).
--- Update the paths below to match the actual location on the DB server.
+-- The COPY commands below require the txt files to exist on the DB server filesystem with appropriate permissions.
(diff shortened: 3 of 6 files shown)


libsys_airflow/plugins/data_exports/marc/exporter.py
```diff
@@ -210,7 +210,6 @@ def write_marc(
         logger.info(f"Writing to directory: {directory}")
         directory.mkdir(parents=True, exist_ok=True)
         marc_file = directory / f"{marc_file_name}.mrc"
-        marc_file.touch()
```
shelleydoljack, line 213: Why is this no longer needed?

libsys_airflow/plugins/data_exports/sql/hrids_mat_view.sql
```diff
@@ -0,0 +1,35 @@
+-- Materialized view built from flat lists of FOLIO instance hrids.
+-- The COPY commands below require the txt files to exist on the DB server filesystem with appropriate permissions.
+-- You can have more than one hrid file, but make sure the txt files as a group contain only unique hrids, one per line, and no header row.
+
+DROP TABLE IF EXISTS public.hrid_export_list;
+CREATE TABLE public.hrid_export_list (hrid text);
+COPY public.hrid_export_list FROM '/home/folio/hrids.txt';
+-- COPY public.hrid_export_list FROM '/home/folio/hrids2.txt';
```
jermnelson, line 8: Kruft?

jermnelson (approved):
Approved but I think there is a leftover line that commented out from development.

.gitignore
```diff
@@ -20,3 +20,5 @@ vendor-keys/
 digital-bookplates/
 authorities/
 sdr-files/
+libsys_airflow/plugins/data_exports/sql/*
+!libsys_airflow/plugins/data_exports/sql/**/*.sql
```
shelleydoljack: Is this cruft? Not sure why it's here.

libsys_airflow/plugins/data_exports/sql/hrids_mat_view.sql
jgreben: I actually left that in on purpose to show that you could have multiple files there if needed.

libsys_airflow/plugins/data_exports/marc/exporter.py
jgreben: I think it's coming into play with the upgraded `s3transfer` package.
```
The s3path library's touch() calls exists() internally, and s3path's exists() implementation uses ListObjectVersions (which requires s3:ListBucketVersions) rather than a simpler HeadObject check.

The root cause is a mismatch between what s3path requires and what the IAM user folio-app-dev is allowed to do in the dev environment. The s3:ListBucketVersions permission is not granted.
```

.gitignore
jgreben: So that the sql directory only has sql in it. Locally I have some of those txt files and I wanted to ignore those. If you want I can use my `.gitignore_global` instead.

libsys_airflow/plugins/data_exports/marc/exporter.py
shelleydoljack: Can you test whether data exports that don't use S3 work without this marc_file.touch()? Otherwise, maybe we need to ask Operations to add that permission to the IAM users for folio.

jgreben: The test passes at https://github.com/sul-dlss/libsys-airflow/blob/gb-hrid-dump/tests/data_exports/test_marc_exports.py#L185

This was successful: https://sul-libsys-airflow-dev.stanford.edu/dags/select_pod_records/runs/manual__2026-06-05T23:25:10.780610+00:00/tasks/fetch_marc_records_from_folio

.gitignore
shelleydoljack: Yes, please use your local gitignore.

PR #1830, opened 2026-07-14
Adds selection dates for data export instead of dag run-time variables.

### LINKED ISSUE #1814: Check for DAGs that use runtime-variable values in Dag construction.
In Airflow3 DAGs with variables may be showing the error:
```
This Dag uses runtime-variable values in Dag construction.
It causes the Dag version to increase as values change on every Dag parse.
```

For example in the pod_Selections dag the datetime.now() calls are the primary culprit since they change every parse cycle. The fix is to remove the dynamic defaults and either use a static placeholder (e.g., "" or "yesterday") or compute the dates inside the task callable instead.

shelleydoljack:
Closes #1814

### DIFF
--- a/libsys_airflow/plugins/data_exports/instance_ids.py
+++ b/libsys_airflow/plugins/data_exports/instance_ids.py
@@ -21,12 +21,6 @@ def _vendor_selection_dates(vendor: str) -> tuple:
         case "gobi":
             from_date = f"{(datetime.now(pacific_timezone) - timedelta(8)).strftime('%Y-%m-%d')}"
             to_date = f"{(datetime.now(pacific_timezone)).strftime('%Y-%m-%d')}"
-        case "oclc":
-            from_date = f"{(datetime.now(pacific_timezone) - timedelta(1)).strftime('%Y-%m-%d')}"
-            to_date = f"{(datetime.now(pacific_timezone)).strftime('%Y-%m-%d')}"
-        case "pod":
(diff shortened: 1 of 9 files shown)


libsys_airflow/plugins/data_exports/instance_ids.py
```diff
@@ -2,13 +2,36 @@
...
 from typing import Union
 
 from airflow.sdk import get_current_context
 from airflow.providers.common.sql.operators.sql import SQLExecuteQueryOperator
 
 logger = logging.getLogger(__name__)
+pacific_timezone = ZoneInfo("America/Los_Angeles")
+
+
+def _vendor_selection_dates(vendor: str) -> tuple:
```
jgreben, line 16: Maybe DRY up `to_date = f"{(datetime.now(pacific_timezone)).strftime('%Y-%m-%d')}"` by setting it first? Also is backstage supposed to be different (there's probably a good reason and if you happen to know it maybe add a comment about it)?

jgreben (approved):
Non-blocking suggestions.

libsys_airflow/plugins/data_exports/instance_ids.py
jgreben: Oh yes, now I remember... Backstage is weekly so we give it a "buffer" day.

shelleydoljack: @jgreben How about this? https://github.com/sul-dlss/libsys-airflow/pull/1830/commits/901fa68083319c67872c5c20d122bc1234973721

jgreben (approved):

